Issue #12, opened 2016-09-10
Add support for other DataObjects versions


synaptic:
I merged the compression branch into master and it seemed to work fine in reading some compressed HDF5 files I had.  However, when I grabbed some older files which were created by a different codebase, I got an [AssertError](https://github.com/jjhelmus/pyfive/blob/master/pyfive/low_level.py#L625) from `DataObjects.get_data()` saying that the version was not equal to 3.  Are you going to add support for older dataset versions (1 and 2)?  If not, I can try to read the spec and attempt to extend your framework to those versions.

jjhelmus:
@synaptic You are correct, pyfive only supports version 3 data layout messages at the moment. The layout of all three version is specified in [The Data Layout Message](https://www.hdfgroup.org/HDF5/doc/H5.format.html#LayoutMessage) section of the HDF5 specification.  This message used to be title "Data Storage" which explains the names in pyfive.  Version 1 and 2 were used in version 1.4 and 1.6.x of the HDF5 library.  Generating sample files to tests these older versions may be challenging with more modern versions of the library.

I'd be happy to include support for dataset layout messages version 1 and 2 in pyfive but this is not a priority for me as I don't any files which use these version.  I'd be happy to review a PR that did add support.

kmuehlbauer:
Looks like this is fixed by #16.

PR #16, opened 2016-10-10
Add support for v1 and v2 data objects

synaptic:
I have tested it with the committed file (v1 data object which is compressed) as well as many of my own larger files.  I have not tested any v2 data objects since I don't have any.  But it should work per the HDF5 spec.  All the other unit tests pass.


synaptic:
Forgot to mention, this will fix #12

jjhelmus:
Really nice addition @synaptic, thanks for this!

PR #42, opened 2019-03-22
property offset is the same for layout_class 1 and 2

bmaranville:
within versions 1 and 2 of data layout message: see https://support.hdfgroup.org/HDF5/doc/H5.format.html#LayoutMessage

With this change, should be able to load data from test file with version 2 data layout message, and layout class==1
https://bitbucket.hdfgroup.org/projects/HDFFV/repos/hdf5/browse/test/filespace_1_6.h5


jjhelmus:
Yes there should be no difference in the reserve size between layout class 1 and 2, thanks for the fix.

PR #43, opened 2020-05-16
adding an implementation of SoftLink

bmaranville:
This adds an implementation of resolving SoftLinks.  The variety of SoftLink that appears in SymbolTable.get_links has been tested and works, but I couldn't find any files that contain link messages (0x0002) in the tests or in my typical working files, so that has not been tested (the _get_links_from_link_msgs method in DataObjects)

Implementation notes: soft links are distinguished from hard links by the fact that they are strings, in the 'get' method in high_level.Group class.  If a link is a string (and therefore a soft link) it is sent back to "get" as a string, and if the lookup fails None is returned, as in the spec.

### DIFF
--- a/pyfive/dataobjects.py
+++ b/pyfive/dataobjects.py
@@ -501,7 +501,7 @@ class DataObjects(object):
         for symbol_table_address in btree.symbol_table_addresses():
             table = SymbolTable(self.fh, symbol_table_address)
             table.assign_name(heap)
-            links.update(table.get_links())
+            links.update(table.get_links(heap))
         return links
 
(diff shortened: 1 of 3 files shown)


bmaranville:
It looks like it's only failing because Python3.3 is not available for testing anymore, and it's still on the list of targets.

jjhelmus:
@bmaranville This looks great, thanks for putting this together.  I'll fix up the CI in another PR, there is no need to test on Python 3.3 (or even 3.4) and 3.7 and 3.8 should be added.

PR #48, opened 2021-07-19
support for V2 filter pipeline descriptions

bmaranville:
I've been seeing files (e.g. created by netCDF4 library) that are using version 2 of the filter pipeline description, as described at https://support.hdfgroup.org/HDF5/doc/H5.format.html#FilterMessage

This patch enables support.


woutdenolf:
Do you have an example file for the test suite?

bmaranville:
Superseded by #52

PR #49, opened 2021-10-11
adding a test for filter pipeline description v2

bmaranville:
Here is a test for filter pipeline description version 2, using [a file from the netCDF4 examples](https://github.com/Unidata/netcdf4-python/blob/master/examples/data/prmsl.2011.nc)

### DIFF
--- a/tests/test_filter_pipeline_v2.py
+++ b/tests/test_filter_pipeline_v2.py
@@ -1,27 +1,19 @@
 """ Unit tests for pyfive's ability to read a file with filter_pipeline v2
 (as is found in some new netCDF4 files) """
 import os
-import warnings
 
-import numpy as np
-from numpy.testing import assert_array_equal, assert_almost_equal
+from numpy.testing import assert_almost_equal
 
 import pyfive
 
 DIRNAME = os.path.dirname(__file__)
-# test file from: https://github.com/Unidata/netcdf4-python/blob/master/examples/data/prmsl.2011.nc
-NETCDF4_FILE = os.path.join(DIRNAME, 'prmsl.2011.nc')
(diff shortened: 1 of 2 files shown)


tests/test_filter_pipeline_v2.py
```diff
@@ -0,0 +1,27 @@
+""" Unit tests for pyfive's ability to read a file with filter_pipeline v2
+(as is found in some new netCDF4 files) """
+import os
+import warnings
```
kalvdans: Unused import, remove

tests/test_filter_pipeline_v2.py
```diff
@@ -0,0 +1,27 @@
...
+
+import numpy as np
+from numpy.testing import assert_array_equal, assert_almost_equal
```
kalvdans: Unused imports, remove

bmaranville: > 8 MB is a bit on the large side, but better than nothing.

You're right - if I can find a smaller example file that uses the new filter pipeline description I will make a new pull request with that instead.  I tried for a while to create a new file using netCDF4 but was unable to trigger the switch to filter pipeline description v2.

kmuehlbauer: @bmaranville 

> I tried for a while to create a new file using netCDF4 but was unable to trigger the switch to filter pipeline description v2.

That would be the way forward, I think.

bmaranville:
I was able to create a small file with filter pipeline description version 2.  I will manually merge this with the feature branch and create a new PR with both the feature and test.

kmuehlbauer:
@bmaranville That sounds great :+1:

bmaranville:
Superseded by #52

PR #52, opened 2022-05-11
Filter pipeline description v2

bmaranville:
I've been seeing files (e.g. created by netCDF4 library) that are using version 2 of the filter pipeline description, as described at https://support.hdfgroup.org/HDF5/doc/H5.format.html#FilterMessage

This patch enables support. (replaces earlier PRs #48 and #49)


kmuehlbauer:
@bmaranville  I'll try to review this the next day. It looks like Travis CI isn't run any more. How do you feel about moving CI to GitHub Actions?

pyfive/dataobjects.py
```diff
@@ -347,33 +347,61 @@ def filter_pipeline(self):
...
-            client_data = struct.unpack_from(fmt, self.msg_data, offset)
-            filter_info['client_data'] = client_data
-            offset += 4 * filter_info['client_data_values']
-
-            if filter_info['client_data_values'] % 2:
-                offset += 4  # odd number of client data values padded
-
-            filters.append(filter_info)
+        if version == 1:
+            res0, res1 = struct.unpack_from('<HI', self.msg_data, offset)
```
kmuehlbauer, line 353: Just a side note, to not forget:

Would it be more convenient, if we move unpacking/offset-increment into a function?

We might also think about using a given structure (eg dict), which has the bits and pieces (name, unpacking information etc), to iterate over the `msg_data` and automatically unpack.

kmuehlbauer: Also wondering what `res0`, `res1` is in this context (probably need to immerse myself into HDF5 format). It is not needed anywhere else in this function.

kmuehlbauer (approved):
@bmaranville This is looking good to me. I've added some comments which are not connected to the changes in this PR to not forget about that.

pyfive/dataobjects.py
bmaranville: ```res0``` and ```res1``` are nothing exciting... they are reserved sections and expected to contain zeros: https://docs.hdfgroup.org/hdf5/develop/_f_m_t3.html#FilterMessage

bmaranville: I would agree that streamlining unpacking and moving the offset would be a useful thing to do - it is largely manual and error-prone (what if you forget to add the offset advance!).  There are a few places where you read but don't advance, but those could still be handled manually.

kmuehlbauer:
:tada: @bmaranville I hope to use this enhancement soon.

PR #55, opened 2024-02-22
Minimal change to avoid a numpy deprecation failure. Without this

bnlawrence:
All tests fail at python 3.10 and numpy 1.24.4. With it, they all pass.


kmuehlbauer:
@bnlawrence Shall the build matrix be updated, here? (with python 3.11, python 3.12)

kalvdans:
Remove your play code from the pull request. Post it as comments intead.

bnlawrence:
Oh, sorry, I forgot that it would end up in the pull request because I did it on master. My bad. Will get that done later today (in meetings now).

bnlawrence:
Ok, I think that's clean on master now. Sorry about the noise.

bnlawrence:
> @bnlawrence Shall the build matrix be updated, here? (with python 3.11, python 3.12)

I've not tested it on them yet. If we can do what we're trying to do, we will need to test it properly, and we can get back to that.

kmuehlbauer (approved):

jjhelmus:
LGTM, thanks for the fix.

jjhelmus:
> Shall the build matrix be updated, here? (with python 3.11, python 3.12)
Including newer Python versions in the test matrix is a good idea. I'll submit a PR adjusting this later today.

PR #56, opened 2024-02-22
Retrieve maxshape for Dataset

### LINKED ISSUE #50: maxshape-attribute missing on datasets
See https://docs.h5py.org/en/stable/high/dataset.html?highlight=maxshape#h5py.Dataset.maxshape

bmaranville:
retrieving maxshape from dataobjects and exposing maxshape property on Dataset

closes #50

### DIFF
--- a/pyfive/dataobjects.py
+++ b/pyfive/dataobjects.py
@@ -246,7 +246,16 @@ class DataObjects(object):
         """ Shape of the dataset. """
         msg = self.find_msg_type(DATASPACE_MSG_TYPE)[0]
         msg_offset = msg['offset_to_message']
-        return determine_data_shape(self.msg_data, msg_offset)
+        shape, maxshape = determine_data_shape(self.msg_data, msg_offset)
+        return shape
+
+    @property
+    def maxshape(self):
+        """ Maximum Shape of the dataset. (None for unlimited dimension) """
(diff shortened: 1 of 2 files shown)


pyfive/dataobjects.py
```diff
@@ -648,6 +657,8 @@ def is_dataset(self):
         """ True when DataObjects points to a dataset, False for a group. """
         return len(self.find_msg_type(DATASPACE_MSG_TYPE)) > 0
 
+UNLIMITED_SIZE = struct.unpack('<Q', b'\xff\xff\xff\xff\xff\xff\xff\xff')[0]
```
kalvdans: easier written `(1 << 64)  - 1`

bmaranville: Fair enough :) this was copy-paste from some earlier code, which has since been refactored to be 
```python
UNDEFINED_ADDRESS = 0xffffffffffffffff
```
(https://github.com/jjhelmus/pyfive/blob/master/pyfive/core.py#L10), which could also be re-used here as a constant.

jjhelmus:
The use of `determine_data_shape` on line 183 will also need to be updated with the new return type

jjhelmus:
LGTM, thanks for the PR

jjhelmus (approved):

PR #57, opened 2024-02-22
add Python 3.11 and 3.12 to test matrix in CI

jjhelmus:


kmuehlbauer (approved):
Great, working without flaws!

PR #61, opened 2024-04-26
Adding support for reading only chunks and various pieces of the H5Py lower level interface

bnlawrence:
The key changes are the introduction of a subclass of DataObjects which represents a singleton variable DataObject, and a new H5D class which provides the lower level chunk interface. Arguably these two classes could be combined.


bnlawrence:
Argh Sorry. This was supposed to be going into my main. For checking before making a pull request here.

Issue #64, opened 2024-12-21
gzip compression options not visible via dataset.compression_opts with v2 filter pipelines


bnlawrence:
This issue is exposed by adding a small test to `test_filter_pipeline_v2`

```
def test_filter_pipeline_compression_opts_v2():

     with pyfive.File(FILTER_PIPELINE_V2_FILE) as hfile:
        assert 'data' in hfile
        d = hfile['data']
        # the point of this test is to ensure we can actually retrieve the compression opts
        x = d.compression_opts
```
This will fail with 
```
  @property
    def compression_opts(self):
        """ Compression filter options, None is no options/compression. """
        if self._filter_ids is None:
            return None
        if GZIP_DEFLATE_FILTER in self._filter_ids:
            gzip_entry = [d for d in self.filter_pipeline
                          if d['filter_id'] == GZIP_DEFLATE_FILTER][0]
>           return gzip_entry['client_data'][0]
E           KeyError: 'client_data'

pyfive/dataobjects.py:324: KeyError
```
which I think happens because the v2 spec has `client_data_values` not `client_data` in this dictionary.

bnlawrence:
So for example, this is what gzip_entry looks like in test_compressed_dataset:
```
odict_keys(['filter_id', 'name_length', 'flags', 'client_data_values', 'filter_name', 'client_data'])
```
and this is what it looks like in the test above:
```
odict_keys(['filter_id', 'optional', 'name', 'client_data_values'])
```

bnlawrence:
Well, for now, this seems to work, but I don't really know what is going on (well, I know what the option is, but the spec [here]() implies that the filter description should not differ between v1 and v2.

```
key = {0:'client_data_values',1:'client_data'}['client_data' in gzip_entry]
return gzip_entry[key][0
```

PR #65, opened 2025-01-07
Issue64

bnlawrence:
This is the simple fix described for #64, includes a test that exposes it, and the one liner to fix it. Works for a number of other data files.


bnlawrence:
Oh well, that's embarrassing, I see the comments in my commits got the issue number wrong. This is why I need someone in my team to review things before I am allowed to do anything in public. Will redo.

PR #66, opened 2025-01-07
Issue64

bnlawrence:
This is the simple fix described for #64, includes a test that exposes it, and the one liner to fix it. Works for a number of other data files. Now includes the correct issue number in the commits.

### DIFF
--- a/pyfive/dataobjects.py
+++ b/pyfive/dataobjects.py
@@ -321,7 +321,8 @@ class DataObjects(object):
         if GZIP_DEFLATE_FILTER in self._filter_ids:
             gzip_entry = [d for d in self.filter_pipeline
                           if d['filter_id'] == GZIP_DEFLATE_FILTER][0]
-            return gzip_entry['client_data'][0]
+            key = {0:'client_data_values',1:'client_data'}['client_data' in gzip_entry]
+            return gzip_entry[key][0]
         return None
 
     @property
--- a/tests/test_filter_pipeline_v2.py
+++ b/tests/test_filter_pipeline_v2.py
@@ -26,3 +26,4 @@ def test_filter_pipeline_compression_opts_v2():
         d = hfile['data']
         # the point of this test is to ensure we can actually retrieve the compression opts
         x = d.compression_opts
+        assert x == 9
--- a/tests/test_filter_pipeline_v2.py
+++ b/tests/test_filter_pipeline_v2.py
@@ -17,3 +17,12 @@ def test_filter_pipeline_descr_v2():
         d = hfile['data']
         assert d.shape == (10,10,10)
         assert_almost_equal(d[0,0,0], 1.0)
+
+
+def test_filter_pipeline_compression_opts_v2():
+
+     with pyfive.File(FILTER_PIPELINE_V2_FILE) as hfile:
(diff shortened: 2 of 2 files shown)


tests/test_filter_pipeline_v2.py
```diff
@@ -17,3 +17,12 @@ def test_filter_pipeline_descr_v2():
...
         assert_almost_equal(d[0,0,0], 1.0)
+
+
+def test_filter_pipeline_compression_opts_v2():
+
+     with pyfive.File(FILTER_PIPELINE_V2_FILE) as hfile:
+        assert 'data' in hfile
+        d = hfile['data']
+        # the point of this test is to ensure we can actually retrieve the compression opts
+        x = d.compression_opts
```
kalvdans: As an enhancement to the unit test, could you also check the value of  `x`  to be correct?

pyfive/dataobjects.py
```diff
@@ -321,7 +321,8 @@ def compression_opts(self):
         if GZIP_DEFLATE_FILTER in self._filter_ids:
             gzip_entry = [d for d in self.filter_pipeline
                           if d['filter_id'] == GZIP_DEFLATE_FILTER][0]
-            return gzip_entry['client_data'][0]
+            key = {0:'client_data_values',1:'client_data'}['client_data' in gzip_entry]
+            return gzip_entry[key][0]
         return None
```
kalvdans, line 326: By the documentation pointed to in your issue, [[link](https://davis.lbl.gov/Manuals/HDF5-1.8.7/H5.format.html#FilterMessage)], it seems there is only a single list of client data values. Instead of your suggested fix, I think we instead should search-and-replace in the code and never produce any dictionary with `"client_data"` as the key. (disclaimer: I haven't actually studied the source code yet)

bnlawrence: I can see arguments for doing it both ways, the code I've committed is as faithful to what is in the file at point of use, to do different would require hacking the message structure before using it ... I'd defer to the package owners on what is the best option.

kalvdans: That's fine, thanks for elaborating!

bnlawrence: (At some point you'd have to do what I've got above, insofar as we see files with both compression option structures, we can't control what we will get to have to read.)

bmaranville:
I think there is a bug in `dataobjects.py`, when reading the filter pipeline.

For version 1 pipeline descriptions, `filter_info["client_data_values"]` is read directly from the message and corresponds to the *length* of the client data array.  The values are then read into `filter_info["client_data"]`.

For version 2 pipelines, the *values* in the client data array are stored in `filter_info["client_data_values"]`, and a different variable (`num_client_values`) is used to temporarily store the length of the array.

This is incorrect - the values should be stored in `filter_info["client_data"]` as they are for v1, and possibly for consistency we could store the length of the value array in `filter_info["client_data_values"]` (though that value would probably not get used again)

If this is fixed then the changes to `dataobjects.py` above are not needed.

bnlawrence:
Yes, it works for me too! What's the best way forward? Do you want me to fix this in my pull request and go from there?

bnlawrence:
This was overtaken by events

PR #67, opened 2025-01-07
Empty unlimited chunked variables cause crash

bnlawrence:
If ones has created a variable which is intended to be chunked, but it is currently empty, when the file is read, we get a stack dump that ends with this:
```
    def _read_node_header(self, offset, node_level):
        """ Return a single node header in the b-tree located at a give offset. """
>       self.fh.seek(offset)
E       ValueError: cannot fit 'int' into an offset-sized integer
```

This pull request includes code to create a file which manifests the problem, a test to expose it, and a fix.

### DIFF
--- a/pyfive/dataobjects.py
+++ b/pyfive/dataobjects.py
@@ -480,6 +480,10 @@ class DataObjects(object):
 
     def _get_chunked_data(self, offset):
         """ Return data which is chunked. """
+
+        if np.prod(self.shape) == 0:
+            return np.empty(self.shape, dtype=self.dtype)
+
         self._get_chunk_params()
         chunk_btree = BTreeV1RawDataChunks(
             self.fh, self._chunk_address, self._chunk_dims)
--- a/tests/make_netcdf_unlimited.py
+++ b/tests/make_netcdf_unlimited.py
@@ -1,0 +1,18 @@
+#! /usr/bin/env python
+""" Create a netcdf file with an unlimited dimension, but no data """
+
+import netCDF4
+import numpy as np
+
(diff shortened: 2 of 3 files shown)


pyfive/dataobjects.py
```diff
@@ -480,6 +480,10 @@ def _get_contiguous_data(self, property_offset):
 
     def _get_chunked_data(self, offset):
         """ Return data which is chunked. """
+
```
bmaranville, line 483: I think this will work - it is also possible to test if the chunk address is `UNDEFINED_ADDRESS`, which will happen when no data has been written to the Dataset yet (see change in https://github.com/usnistgov/jsfive/commit/f2284207c5fb3d8c6abe68864e7c757a205e92df , which I should have backported to pyfive)

EDIT - I think maybe the test for `UNDEFINED_ADDRESS` is important here because sometimes you encounter datasets with non-zero shapes but which have not been written yet (initializing a dataset and writing data to it are two separate steps).

bnlawrence: That seems sensible. At this point I'm not minded to follow through fixing this here, as where it gets done in the new H5D.py will be slightly different. What I have there is:
```
# look out for an empty dataset, which will have no btree
if np.prod(self.shape) == 0 or dataobject._chunk_address == UNDEFINED_ADDRESS:
    self._index = {}
    return
```
(This is in the context of caching the b-tree when we instantiate a DatasetID, which we do when we create a variable instance with eg. `x=myfile['variable']. We do that at this point so that all threads in a thread pool have their b-tree before they get going on their bit of work.)

valeriupredoi:
@bnlawrence @davidhassell is this still relevant in v0.5.0? :beer:

Issue #41, opened 2018-02-27
unable to read attributes when the size of attribute lists is relatively big


jduan8109:

jjhelmus:
@jduan8109 Can you provide additional details or an example of this behavior?  Either would make debugging the issue much easier.

jduan8109:
@jjhelmus , thanks. I am trying to load a netcdf4 file in with your code. When i have a short list of attributes in the data like this:
//group attributes
		:parent_grid_name = "GERMANY" ;
		:number_of_subgrids = "0" ;
	        :minimum_latitude = "50.85" ;
		:maximum_latitude = "51.7" ;
		:left_longitude = "12.08333333333333" ;
		:right_longitude = "13.5" ;
I have no problem getting all the attributes via calling "xxx.attrs", however, if there is a long list of attributes in the file like below, using "xxx.attrs" will return an empty dict instead.

//group attributes
		:parent_grid_name = "GERMANY" ;
		:number_of_subgrids = "0" ;
	        :minimum_latitude = "50.85" ;
		:maximum_latitude = "51.7" ;
		:left_longitude = "12.08333333333333" ;
		:right_longitude = "13.5" ;
		:north_south_node_interval = "0.001666666666666667" ;
		:east_west_node_interval = "0.08333333333333333" ;
		:north_south_node_sequence = "south_to_north" ;
                :east_west_node_sequence = "west_to_east" ;
		:array_node_order_arrangement_method = "row_major" ;
	   :number_of_rows = "766" ;
	   :number_of_columns = "851" ;
	   :recommended_interpolation_method = "bilinear" ;
	   :dataset_date_time = "13-01-03" ;
	   :dataset_begin_date = "13-01-03" ;
	   :dataset_end_date = "NONE" ;
	   :dataset_epoch = "2013.163" ;

bnlawrence:
@jduan8109 I don't think I've seen this, and I'm in the process of providing some improvements to pyfive. If you're still interested, and it's still a problem, could I trouble you to create a little netcdf file with just the attributes which cause the problem, and upload it here?

bnlawrence:
@jduan8109 we don't need more details! One of my colleagues has just exposed what I now believe to be the same issue (https://github.com/NCAS-CMS/pyfive/issues/23). While he found it using our fork, I've tested it with the current master from here, and his 'B file" also returns an empty attribute set ... 

[...]

Issue #60, opened 2024-03-07
pyfive doesn't play nicely with s3fs


bnlawrence:
This is probably a problem with s3fs, but for now it's worth noting here. In practice, we cannot read contigous data from a file opened on an s3fs. Here's some toy code to show the problem:
```python
import s3fs
import pyfive as p5
from pathlib import Path
import numpy as np

S3_URL = "redacted"
S3_BUCKET = 'redacted'

def local(fname,v):

    mypath = Path(__file__).parent
    f = p5.File(mypath/fname)
    d = f[v]
    dd = d[:]
    return dd

def s3(fname, v):
        
    fs = s3fs.S3FileSystem(anon=True, client_kwargs={'endpoint_url': S3_URL})
    uri = S3_BUCKET + '/' + fname

    with fs.open(uri,'rb') as s3file2:
        f2 = p5.File(s3file2)
        d = f2[v]
        dd = d[:]
        return dd

if __name__=="__main__":
    fn, v = 'common_cl_a.nc', 'cl'
    d1 = local(fn,v)
    d2 = s3(fn,v)
    np.testing.assert_array_equal(d1,d2)
```
which gives:
```
Traceback (most recent call last):
  File "/Users/bnl28/Repositories/pyfive/bnl/mytest_s3_etc.py", line 32, in <module>
    d2 = s3(fn,v)
  File "/Users/bnl28/Repositories/pyfive/bnl/mytest_s3_etc.py", line 25, in s3
    dd = d[:]
  File "/Users/bnl28/Repositories/pyfive/pyfive/high_level.py", line 279, in __getitem__
    data = self._dataobjects.get_data(args)
  File "/Users/bnl28/Repositories/pyfive/pyfive/dataobjects.py", line 630, in get_data
    return self._get_contiguous_data(self.property_offset)[args]
  File "/Users/bnl28/Repositories/pyfive/pyfive/dataobjects.py", line 671, in _get_contiguous_data
    return np.memmap(self.fh, dtype=self.dtype, mode='c',
  File "/Users/bnl28/mambaforge/envs/fesom/lib/python3.10/site-packages/numpy/core/memmap.py", line 267, in __new__
    mm = mmap.mmap(fid.fileno(), bytes, access=acc, offset=start)
io.UnsupportedOperation: fileno
```
[...]

Issue #59, opened 2024-03-04
dataset dtype is not a numpy dtype


bnlawrence:
I believe the HDF5 Dataset dtype is a numpy dtype (and hence has attributes like itemsize).

The easiest fix to this would be to return a numpy dtype in the Dataset property, but this would also require a change to the `test_repr`. Happy to make a pull request on this if desired.

kmuehlbauer:
I'd follow the README and consider that a bug which should be fixed.

bnlawrence:
(This has been hoovered up in the h5netcdf support to be incoming in the next couple of weeks.)

jjhelmus:
Confirming this as a bug. h5py returns a NumPy dtype object for the `dtype` attribute of a dataset, pyfive should as well.

bnlawrence:
I believe this was fixed in the recent updates.

PR #68, opened 2025-01-30
Functionality enhancements to address lazy loading of chunked data, variable length strings, and other minor bug fixes

bnlawrence:
This pull request was originally prompted by #6, insofar as we needed lazy loading of chunked data, but also our need to a) have a pure python backend reader for `h5netcdf`, and b) which could expose variable b-trees to other downstream software. 

Our motivations included thread-safety and performance at scale in a cloud environment. To do this we have implemented versions of some more components of the `h5py` stack, and in particular, a version of the `h5d.DatasetID` class, which is now holds all the code which is used for data access (as opposed to attribute access, which still lives in `dataobjects`).  There are a couple of extra methods for exposing the chunk index directly rather than via an iterator and to access chunk info using the zarr indexing scheme rather than the `h5py` indexing scheme. 

The code also includes an implementation of what we have called `pseudochunking` which is used for accessing a contiguous array which is larger than memory via S3.  In essence all this does is declare default chunks aligned with the array order on disk and use them for data access.

There are many small bug fixes and optimisations to support cloud usage, the most important of which is that once a variable is instantiated (i.e. for an open `pyfive.File` instance `f`, when you do `v=f['variable_name']`) the attributes and b-tree are read, and it is then possible to close the parent file (`f`), but continue to use (`v`) - and we have test coverage that shows that this usage of `v` is thread-safe (there is a test which demonstrates this, it's slow, but it needs to be as shorter tests were sporadically passing). (The test harness now includes all the components necessary for testing pyfive accessing data via both Posix and S3).

As well as closing #6, this pull request would close: #41,#59,#60,#64.


bnlawrence:
(I see we're failing the checks due to some build dependencies. Will sort that shortly.

valeriupredoi:
> (I see we're failing the checks due to some build dependencies. Will sort that shortly.

I've switched to `pip install .[test]` in the GA workflow so we get them deps for testing (only) :+1:

.gitignore
```diff
@@ -1,2 +1,10 @@
 .coverage
+.pyc
 build
+__pycache__/
+*.egg-info
+.idea
+.DS_Store
+test-reports/
+<_io.Bytes*>
```
kalvdans, line 9: If filenames with < in them are generated, I'd like to see them.

valeriupredoi (changes requested):
@bnlawrence you got three instances of `match case:` - that is Python>=3.10 syntax, see https://github.com/jjhelmus/pyfive/actions/runs/13996247036/job/39192033797

I would argue you just drop support for Python 3.9 since 3.9 it will be fully retired in August anyway :beer:

bnlawrence:
Hi Folks. We're rather desperate to announce this to the community, and I want to do that via a poster, which I will have to send for printing on Wednesday. Ideally we have links to this repository and the main ... what do we need to do to get this accepted? The code has been reviewed by three of us here at NCAS (and we will continue to maintain it).  We'd prefer not to make changes to support Python 3.9 as that's only got months to live, and we think that few people will be using this in the wild.

valeriupredoi:
hi folks, just a heads up that @jjhelmus and @bnlawrence had a very constructive conversation, and the decision was taken to allow the move of this repo over to our NCAS-CMS organization https://github.com/NCAS-CMS where we will be taking very good care of Pyfive. I am inclined to have this merged before we do the move, and also, would like to ask the developers and contributors here if they'd be happy for me to add them to the moved repo (once we move it, and if the move missed any of them). Many thanks again @jjhelmus :beer:

valeriupredoi:
thanks for looking and confirming @kmuehlbauer and @jjhelmus - I am starting the operations now: first step is to retire support for the defunct Python 3.9, so we can merge this PR with no test failures (and obviously, repair anything may be caused by a new generation of the environment here) :beers:

Issue #51, opened 2022-01-24
Project Status


kmuehlbauer:
@jjhelmus

Could you please give a short status on the project?

It seems there is at least some interest to enhance and improve `pyfive`.

Xref: 
- https://github.com/jjhelmus/pyfive/pull/47#issuecomment-881450826
- https://github.com/jjhelmus/pyfive/pull/47#issuecomment-939929311

Ping @bmaranville @woutdenolf

It would be great if there would be a way forward for `pyfive`.

`pyfive` is on the wanted list as [`h5netcdf` backend quite a while](https://github.com/h5netcdf/h5netcdf/issues/25) and it looks like it has most features already in or in pending PR. 

I would happily join a team of maintainers if one will be established.

jjhelmus:
I have let this project idle for too long, sorry about that.  I no longer work with HDF5 files often so have little need for this library, it no longer is an itch I need to scratch.   

I'd be happy to add additional maintainers who would be willing to triage bugs, review and merge PR and prepare releases.  @kmuehlbauer I've invited you as a collaborator on this repository.

kmuehlbauer:
@jjhelmus Thanks Jonathan, I'll have a look on the current issues and pull requests in the next days. 

@bmaranville @woutdenolf Are you still in the works with HDF5? It would be great if you could join the team on pyfive. It looks like you are much more into the depths of HDF5 format.

bmaranville:
I am indeed still working with HDF5 files regularly... I would be happy to help where I can.

bnlawrence:
So after a long silence, I will be trying to make some significant changes to (a version) of this library, and so I'm keen to understand whether the maintainers of this library would be keen to work with us on those, or whether we should fork and rename. The major changes would be:
- full support for lazy chunk reading
- full support for the H5D.DatasetID read interface (and the use of it by default for Dataset instances).
- a modification to the behaviour of Dataset to make it cache everything it could need so that it can not only support lazy data access, but can operate efficiently in Dask threads (where we want to avoid each thread needing to read the b-tree etc, so we cache the b-tree when we instantiate the new DatasetID class instances).  In practice this means that the Dataset would no longer have a link to the parent dataobjects instance, and utiise only the  DatasetID instance for data access, and the DatasetID would have copied and cached everything it needs from the dataobjects instance). 

There are some new "dummy classes" to support the use of pyfive in the backend of h5netcdf, and some minor bug fixes to core functionality as well,

[...]

PR #69, opened 2025-02-07
Modernize package

valeriupredoi:
Hi @jjhelmus in addition to @bnlawrence's #68 I have put together a PR that is updating the `pyfive` package to fairly modern standards, by:

- adding a conda env file
- pyproject file
- decent GHA workflow
- potential for registering with codecov and displaying cov CI results via that app (in eg PRs)

and a few other small bits and bobs. Please let me know what you think - and many thanks, and a good weekend :beer:


kalvdans:
`pyproject.toml` looks nice and great to remove `setup.py`, moving to a declarative package description instead of a script. :+1:

I'll let others review the other files.

pyproject.toml
```diff
@@ -0,0 +1,105 @@
...
+    "Topic :: Scientific/Engineering :: Physics",
+]
+dynamic = [
+    "readme",
+    "version",
+]
+dependencies = [
+    "numpy",
+    "pip!=21.3",
+    "s3fs",
```
jjhelmus: Is this a run dependency or is this similar to tests? If the later would a `test` extra be a better location for this dependency?

valeriupredoi: hi @jjhelmus it's a run dep needed via #68 see https://github.com/jjhelmus/pyfive/pull/68/files#diff-fa602a8a75dc9dcc92261bac5f533c2a85e34fcceaff63b3a3a81d9acde2fc52 - I can remove it here, and re-add it once if/when #68 gets merged?

bmaranville: `s3fs` is a surprising dependency. Does this have to be installed to use pyfive in the way the majority of users use it (without lazy-loading chunks?)

Also I don't usually see a pip version in the dependencies - is that needed too?

valeriupredoi: nevermind me, I double-checked #68 and `s3fs` is indeed a simple test dep like all the others like `h5netcdf` etc, am gonna move it there right now :+1:

valeriupredoi: done in https://github.com/jjhelmus/pyfive/pull/69/commits/35476b7271414bfb90d227c7b912c7a229d96d98

.github/workflows/pytest.yml
```diff
@@ -1,41 +1,50 @@
...
-        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
-        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
-        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
-    - name: Test with pytest
-      run: |
-        pytest
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: conda-incubator/setup-miniconda@v3
```
kmuehlbauer, line 33: We might be able to reduce the setup time using https://github.com/mamba-org/setup-micromamba. Works quite well over at `h5netcdf` and `xarray`. But I'm fine with the proposed, too.

.github/workflows/pytest.yml
```diff
@@ -1,41 +1,50 @@
...
+        with:
+          activate-environment: pyfive
+          environment-file: environment.yml
+          python-version: ${{ matrix.python-version }}
+          miniforge-version: "latest"
+          use-mamba: true
+      - name: Install Pyfive Test Mode
+        run: |
+          pip install -e .[test]
+      - name: Lint with flake8
```
kmuehlbauer, line 44: Would it make sense to split out linting in a separate job? For the packages I maintain I find it useful to see test breakages even if the linting fails.

kmuehlbauer (approved):
Thanks @valeriupredoi for tackling this. Just two comments from my side.

.github/workflows/pytest.yml
valeriupredoi: that's a good point! We're using setup-miniconda with mamba in a lot of our projects and I know the folks there at conda-incubator too, that's why I used it here too. I reckon we can change to micromamba in the future, if we hit any performance issues :+1:

.github/workflows/pytest.yml
valeriupredoi: that's entirely up to you, folks! We stop and fail the workflow if eg `pre-commit` fails, but I guess that depends how much emphasis one project puts on linting. Eg here we can put a `fail-fast: false` clause so the workflow still fails but it runs all the steps without halting at the first failed step?

valeriupredoi:
many thanks for the review @kmuehlbauer :beer:

.github/workflows/pytest.yml
kmuehlbauer: Yes, sure, my main point is minimizing the CI time. Less time, less energy consumption.

valeriupredoi: same with us, in stark opposition to the Techbro movement :grin: This one's lightning fast too, it'll be even faster once they update the base `mamba` version, for which I already opened an issue :+1:

.github/workflows/pytest.yml
valeriupredoi: I just noticed that the `strategy` already has a `fail-fast: false` so this would mean workflow runs all the steps and marks the wkflow as failed if one step failed, so we good here I reckon :+1:

kmuehlbauer: Great! I must have overlooked the `fail-fast` setting.

.github/workflows/pytest.yml
```diff
@@ -1,41 +1,50 @@
...
 jobs:
   build:
 
+    name: Test Python ${{ matrix.python-version }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
```
jjhelmus: Is there a compelling reason to drop Python 3.9? Is is supported until the end of Oct 2025 (https://endoflife.date/python)

pyproject.toml
```diff
@@ -0,0 +1,105 @@
...
+    "Topic :: Scientific/Engineering :: Atmospheric Science",
+    "Topic :: Scientific/Engineering :: Physics",
+]
+dynamic = [
+    "readme",
+    "version",
+]
+dependencies = [
+    "numpy",
+    "pip!=21.3",
```
jjhelmus: pip should not be included as a package dependency. It is a front end that can be used to install pyfive but it is not needed at runtime.

pyproject.toml
```diff
@@ -0,0 +1,105 @@
...
+]
+description = "A pure python HDF5 reader"
+license = {text = "BSD License, Version 3-Clause"}
+name = "pyfive"
+requires-python = ">=3.10"
```
jjhelmus:
```suggestion
requires-python = ">=3.9"
```

jjhelmus:
I don't think the `python-version` variable is coming into play in the CI with the switch to setup-miniconda. Looking at the test runs they all seem to be running Python 3.12:

For example the [Test Python 3.10 run](https://github.com/jjhelmus/pyfive/actions/runs/13203985085/job/36866023029?pr=69):
```
Run pytest -n 2
============================= test session starts ==============================
platform linux -- Python 3.12.3, pytest-[8](https://github.com/jjhelmus/pyfive/actions/runs/13203985085/job/36866023029?pr=69#step:7:9).3.4, pluggy-1.5.0
rootdir: /home/runner/work/pyfive/pyfive
configfile: pyproject.toml
plugins: html-4.1.1, xdist-3.6.1, metadata-3.1.1, cov-6.0.0
created: 2/2 workers
2 workers [41 items]

.........................................                                [[10](https://github.com/jjhelmus/pyfive/actions/runs/13203985085/job/36866023029?pr=69#step:7:11)0%]
- Generated html report: file:///home/runner/work/pyfive/pyfive/test-reports/report.html -
============================== 41 passed in 0.66s ==============================
```

valeriupredoi:
> I don't think the `python-version` variable is coming into play in the CI with the switch to setup-miniconda. Looking at the test runs they all seem to be running Python 3.12:
> 
> For example the [Test Python 3.10 run](https://github.com/jjhelmus/pyfive/actions/runs/13203985085/job/36866023029?pr=69):
> 
> ```
> Run pytest -n 2
> ============================= test session starts ==============================
> platform linux -- Python 3.12.3, pytest-[8](https://github.com/jjhelmus/pyfive/actions/runs/13203985085/job/36866023029?pr=69#step:7:9).3.4, pluggy-1.5.0
> rootdir: /home/runner/work/pyfive/pyfive
> configfile: pyproject.toml
> plugins: html-4.1.1, xdist-3.6.1, metadata-3.1.1, cov-6.0.0
> created: 2/2 workers
> 2 workers [41 items]
> 
> .........................................                                [[10](https://github.com/jjhelmus/pyfive/actions/runs/13203985085/job/36866023029?pr=69#step:7:11)0%]
> - Generated html report: file:///home/runner/work/pyfive/pyfive/test-reports/report.html -
> ============================== 41 passed in 0.66s ==============================
> ```

It does, up the point where pyfive gets installed in test mode and pytest is run - pytest is installed from PyPI with default base env pip (24) install, rather than the conda env pip (25), also the Python is base env Python. Well spotted @jjhelmus - there seems to be an issue with the cobda env activation, I'll have a look tomorrow, cheers! 🍺

valeriupredoi:
Blast! I know what it is, I forgot to set the default run shell at the top of the workflow - without it it'll always ignore any active environment 😮‍💨

valeriupredoi:
> I don't think the `python-version` variable is coming into play in the CI with the switch to setup-miniconda. Looking at the test runs they all seem to be running Python 3.12:
> 
> For example the [Test Python 3.10 run](https://github.com/jjhelmus/pyfive/actions/runs/13203985085/job/36866023029?pr=69):
> 
> ```
> Run pytest -n 2
> ============================= test session starts ==============================
> platform linux -- Python 3.12.3, pytest-[8](https://github.com/jjhelmus/pyfive/actions/runs/13203985085/job/36866023029?pr=69#step:7:9).3.4, pluggy-1.5.0
> rootdir: /home/runner/work/pyfive/pyfive
> configfile: pyproject.toml
> plugins: html-4.1.1, xdist-3.6.1, metadata-3.1.1, cov-6.0.0
> created: 2/2 workers
> 2 workers [41 items]
> 
> .........................................                                [[10](https://github.com/jjhelmus/pyfive/actions/runs/13203985085/job/36866023029?pr=69#step:7:11)0%]
> - Generated html report: file:///home/runner/work/pyfive/pyfive/test-reports/report.html -
> ============================== 41 passed in 0.66s ==============================
> ```

fixed in https://github.com/jjhelmus/pyfive/pull/69/commits/8de583246614222a1775994176e42b7b8a89210a - that was a sneaky one, great eye for it! :beer:

pyproject.toml
valeriupredoi: I tend to agree with you on this one, but the problem is that it becomes murky which `pip` is used where and by what - we should avoid using `pip` default from PyPI when working with conda packages. If you are 100% that this package will only need `numpy` as a build dependency, then we can remove `pip` here, but if Pyfive will start adding a number of new build dependencies from conda-forge then we should most deffo use the conda-forge `pip` (and have it as a build/install dependency)

.github/workflows/pytest.yml
valeriupredoi: indeed! But a lot of other packages are dropping it like a hot potato - I don't think it's feasible to support it anymore (even if Pyfive has a very small dependency table). Though it's not make or break to me, so I can re-add it, and open an issue to remind us to remove it in eight months :+1:

valeriupredoi:
Python 3.9 re-added in https://github.com/jjhelmus/pyfive/pull/69/commits/32158d4d9c937de8d41ba7ef48af040dd89951e3 and https://github.com/jjhelmus/pyfive/pull/69/commits/f2e9bcbbfcd0f9e1c9ea5f98fbe37fd06ac60ed5 :+1:

pyproject.toml
bmaranville: I'm a huge fan of conda-forge, but pyfive is a pure-Python package and I'm hoping it won't need to be built (or have build dependencies)

.github/workflows/pytest.yml
valeriupredoi: Python 3.9 re-added in https://github.com/jjhelmus/pyfive/pull/69/commits/32158d4d9c937de8d41ba7ef48af040dd89951e3 and https://github.com/jjhelmus/pyfive/pull/69/commits/f2e9bcbbfcd0f9e1c9ea5f98fbe37fd06ac60ed5 :+1:

pyproject.toml
valeriupredoi: sure, but other of its (possible future) dependencies may eg `netCDF4` - getting those off PyPI is a bit of a pain :grin:

jjhelmus: With this listed as a dependency `pip` will be installed into the environment from PyPI by the `pip` installed by `conda` but only if the upgrade strategy is eager (not the default) or something else required a different version of `pip`. In the current run the `conda` installed version is retained.  I can't see any case where this would be needed, the packages listed here are installed by pip itself (or another PEP 517 front end).

kmuehlbauer: > sure, but other of its (possible future) dependencies may eg `netCDF4` - getting those off PyPI is a bit of a pain 😁

I hope, `pyfive` will never depend on `netCDF4` as this depends on `netcdf-c` which depends on `hdf5` which counters the efforts of `pyfive`, IMHO.

valeriupredoi: ah no way, that was a somewhat anecdotal example :smiley: Let me just remobe `pip` from the conda env, then - so that `numpy` is the only one left as install dep :+1:

valeriupredoi:
also relevant to dropping support for Python 3.9 (and 3.10) is [SPEC0](https://scientific-python.org/specs/spec-0000/) if you folks think about subscribing to it :+1:

valeriupredoi:
@jjhelmus what do you reckon, mate - this one's good for merging? :beer:

jjhelmus:
As mentioned in #51, I want to get other involved in this project. I cannot be the only one reviewing and merging PR. Ideally I wouldn't need to do either. 

If someone else can review and merge this PR it would be appreciated.

valeriupredoi:
sounds like a solid plan to me, cheers @jjhelmus :beer: @kmuehlbauer @bmaranville @davidhassell any of you good folk has a minute to have a final review and merge - if all's in order, please? :beers:

pyproject.toml
```diff
@@ -0,0 +1,105 @@
...
+[tool.pytest.ini_options]
+addopts = [
+    "--cov-report=xml:test-reports/coverage.xml",
+    "--cov-report=html:test-reports/coverage_html",
+    "--html=test-reports/report.html",
+]
+
+[tool.coverage.run]
+parallel = true
+source = ["esmvalcore"]
```
kmuehlbauer: `esmvalcore` does not fit here.

kmuehlbauer (changes requested):
I think this PR has been discussed extensively and suggestions have been incorporated. 

I found one more item which needs change, but otherwise this is good to go.

pyproject.toml
valeriupredoi: my bad, inspiration for the file from another of our packages :grin: Fixed in https://github.com/jjhelmus/pyfive/pull/69/commits/62a5b39ddcf967a88dad8cd369570da25dcb6d1c

valeriupredoi:
very many thanks @kmuehlbauer - fixed that little wiggle :beer

kmuehlbauer (approved):

kmuehlbauer:
Thanks @valeriupredoi.

valeriupredoi:
wonderful, great many thanks @kmuehlbauer and everyone else :beers:

PR #73, opened 2025-07-07
PyPI package build GHA minor TLC

valeriupredoi:
Changed OS in pyproject: Independent is not supported anymore; also added our names to the list of package maintainers - @kmuehlbauer would you also want to be added there?


kmuehlbauer:
@valeriupredoi Thanks for inviting me. I'm still committed to help maintaining `pyfive`, but it's a matter of how much of a time I'm able to dedicate. At least for the connection with `h5netcdf` this should work. So, yes, put me on that maintainer list.

valeriupredoi:
@kmuehlbauer wonderful, great many thanks, mate :beers:

Issue #70, opened 2025-06-04
NotImplementedError: Opaque datatype class not supported after last commit


bushaev:
Hey!
I have this piece of code 
```
def extract_offset(path, dset_name="_index"):
    # find index dataset and extract offset
    f = None
    try:
        f = pyfive.File(path)
        dset = f.get(dset_name)
        if dset is None:
            return None
        else:
            return dset._dataobjects.offset

    finally:
        if f is not None:
            f.close()
```
That used to work until the last merged PR, now it gives me this:
```
Traceback (most recent call last):
File "/workspace/cli/jobs/P3_split_audio.py", line 175, in <module>
run_task_cli(
File "/workspace/cli/common/job_utils.py", line 157, in run_task_cli
with artifact_manager_class(
File "/workspace/cli/common/artifacts/artifact_manager.py", line 190, in __exit__
artifact.save()
File "/workspace/cli/common/artifacts/split_audio.py", line 12, in save
make_index(self.h5_path)
File "/workspace/cli/common/artifacts/hdf5_indexer.py", line 91, in make_index
update_offset(path, dset_name)
File "/workspace/cli/common/artifacts/hdf5_indexer.py", line 26, in update_offset
offset = extract_offset(path, dset_name=dset_name)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/workspace/cli/common/artifacts/hdf5_indexer.py", line 14, in extract_offset
dset = f.get(dset_name)
^^^^^^^^^^^^^^^^
File "<frozen _collections_abc>", line 774, in get
File "/customer/.local/lib/python3.11/site-packages/pyfive/high_level.py", line 94, in __getitem__
return Dataset(obj_name, DatasetID(dataobjs), self)
^^^^^^^^^^^^^^^^^^^
File "/customer/.local/lib/python3.11/site-packages/pyfive/h5d.py", line 85, in __init__
if isinstance(dataobject.dtype,tuple):
^^^^^^^^^^^^^^^^
File "/customer/.local/lib/python3.11/site-packages/pyfive/dataobjects.py", line 355, in dtype
return DatatypeMessage(self.msg_data, msg_offset).dtype
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/customer/.local/lib/python3.11/site-packages/pyfive/datatype_msg.py", line 16, in __init__
self.dtype = self.determine_dtype()
[...]

PR #83, opened 2025-08-15
when getting object by address do not fully instantiate when iterating

kmuehlbauer:
## Description

The changes in this PR prevents instantiating objects fully when retrieving objects by address (eg. when dereferencing). This was reported in https://github.com/NCAS-CMS/pyfive/issues/70#issuecomment-3191392693.

## Checklist

- [x] All tests pass

### DIFF
--- a/pyfive/datatype_msg.py
+++ b/pyfive/datatype_msg.py
@@ -39,9 +39,8 @@ class DatatypeMessage(object):
         elif datatype_class == DATATYPE_REFERENCE:
             return ('REFERENCE', datatype_msg['size'])
         elif datatype_class == DATATYPE_ENUMERATED:
-            return ("ENUMERATED", datatype_msg['size'])
-            # raise NotImplementedError(
-            #     "Enumerated datatype class not supported.")
+            raise NotImplementedError(
+                "Enumerated datatype class not supported.")
         elif datatype_class == DATATYPE_ARRAY:
             raise NotImplementedError("Array datatype class not supported.")
         elif datatype_class == DATATYPE_VARIABLE_LENGTH:
--- a/pyfive/datatype_msg.py
+++ b/pyfive/datatype_msg.py
@@ -39,8 +39,9 @@ class DatatypeMessage(object):
         elif datatype_class == DATATYPE_REFERENCE:
             return ('REFERENCE', datatype_msg['size'])
         elif datatype_class == DATATYPE_ENUMERATED:
-            raise NotImplementedError(
-                "Enumerated datatype class not supported.")
+            return ("ENUMERATED", datatype_msg['size'])
+            # raise NotImplementedError(
+            #     "Enumerated datatype class not supported.")
         elif datatype_class == DATATYPE_ARRAY:
             raise NotImplementedError("Array datatype class not supported.")
         elif datatype_class == DATATYPE_VARIABLE_LENGTH:
--- a/pyfive/high_level.py
+++ b/pyfive/high_level.py
@@ -230,7 +230,7 @@ class File(Group):
                 # descend only if it's a subgroup (need to instantiate minimally)
                 # would be neat to have .is_group
                 dataobjs = DataObjects(self.file._fh, link_addr)
-                if not dataobjs.is_dataset and not dataobjs.is_datatype:
+                if dataobjs.is_group:
                     queue.append((full_path, grp[name]))
 
     def close(self):
--- a/pyfive/high_level.py
+++ b/pyfive/high_level.py
@@ -91,7 +91,11 @@ class Group(Mapping):
         if dataobjs.is_dataset:
             if additional_obj != '.':
                 raise KeyError('%s is a dataset, not a group' % (obj_name))
-            return Dataset(obj_name, DatasetID(dataobjs), self)
+            dataset_id = DatasetID(dataobjs)
+            # just leave out enumerated
+            if isinstance(dataset_id.dtype, tuple) and dataset_id.dtype[0] == "ENUMERATED":
+                return None
+            return Dataset(obj_name, dataset_id, self)
        
         try:
             # if true, this may well raise a NotImplementedError, if so, we need
@@ -142,6 +146,9 @@ class Group(Mapping):
         queue = deque(self.values())
         while queue:
             obj = queue.popleft()
+            # skip, when NotImplemented is detected
+            if obj is None:
+                continue
             name = obj.name[root_name_length:]
             ret = func(name, obj)
             if ret is not None:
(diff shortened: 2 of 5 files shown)


kmuehlbauer:
@valeriupredoi I'm really sure there is a better solution to this here. I've only did this branch to let you know, what I'm about to. I've more details in https://github.com/NCAS-CMS/pyfive/issues/70#issuecomment-3191392693.

The main problem is, that pyfive raises `NotImplemented` when a not implemented feature get's instanciated/requested. But, this also happens when dereferencing a Reference using visititems, which will instanciate all variables (including any not implemented) and raise. I can't do anything to prevent or work around from the h5netcdf-side. I hope @bnlawrence can spot the issue and maybe has already an idea to get around that.

bnlawrence:
> @valeriupredoi I'm really sure there is a better solution to this here. I've only did this branch to let you know, what I'm about to. I've more details in [#70 (comment)](https://github.com/NCAS-CMS/pyfive/issues/70#issuecomment-3191392693).
> 
> The main problem is, that pyfive raises `NotImplemented` when a not implemented feature get's instanciated/requested. But, this also happens when dereferencing a Reference using visititems, which will instanciate all variables (including any not implemented) and raise. I can't do anything to prevent or work around from the h5netcdf-side. I hope @bnlawrence can spot the issue and maybe has already an idea to get around that.

I know @davidhassell was thinking about this too. Before I dive in, @davidhassell, did you think any more about it? (As it happens, I am dead sick of proposal spreadsheets, so thinking about this for an hour or two would be a blessed relief.)

kmuehlbauer:
@valeriupredoi Ah, sorry, this should have been a draft.

bnlawrence:
> thanks a lot @kmuehlbauer - mind fixing the test too, please, mate? @bnlawrence you got a minute have a look at Kai's proposed change pls? 🍺

Actually, it would be really useful to have a test that shows the behaviour that is wanted here, which is presumably to fail _when you want it to_ and not before.

kmuehlbauer:
@bmaranville Thanks for chiming in. This seems like a good way forward. I'll try to implement your suggestions here the next day. But, if someone has free cycles, feel free to add to or override here.

pyfive/datatype_msg.py
```diff
@@ -39,8 +39,9 @@ def determine_dtype(self):
         elif datatype_class == DATATYPE_REFERENCE:
             return ('REFERENCE', datatype_msg['size'])
         elif datatype_class == DATATYPE_ENUMERATED:
-            raise NotImplementedError(
-                "Enumerated datatype class not supported.")
+            return ("ENUMERATED", datatype_msg['size'])
```
kmuehlbauer: @bnlawrence So, would this be something useful to return?

bnlawrence: Yes, I think so.  The higher-level API for dtype will need to return uint8, but what we do in between is the key and I think we will need that information.  We could check for that in the higher-level API, return `uint8`, and implement `check_enum_dtype` to look for this.  We could quite quickly then at least implement the ability to look at the enumeration integers, even if we can't immediately decode the dictionary  stored in the Enumeration Property Description.

bmaranville: I think `DATATYPE_ENUMERATED` can have a number of base types (any of the integer types?), not just `uint8` - the base type is specified in the datatype message.

bnlawrence: Quite right. Thanks for the reminder.

bnlawrence:
I have added a new branch ([enum_and_friends](https://github.com/NCAS-CMS/pyfive/tree/enum_and_friends)) where we can work on enum support. For the moment it is very dysfunctional, but it's a framework to work from, and includes a test which should make sure we support the h5py interface. Unfortunately I need to put this down to work on some other things and will be unlikely to follow all the way through before early September - but you never know, depends on progress and boredom levels with the other stuff I *have* to do right now.  Ideally this branch also sorts out the visititems issue properly, but it might be that proceeding with this hack for the h5netcdf application is the right temporary expedient - but someone else will need to write the test for that.

bnlawrence:
> I have added a new branch ([enum_and_friends](https://github.com/NCAS-CMS/pyfive/tree/enum_and_friends)) where we can work on enum support. For the moment it is very dysfunctional, but it's a framework to work from, and includes a test which should make sure we support the h5py interface. Unfortunately I need to put this down to work on some other things and will be unlikely to follow all the way through before early September - but you never know, depends on progress and boredom levels with the other stuff I _have_ to do right now. Ideally this branch also sorts out the visititems issue properly, but it might be that proceeding with this hack for the h5netcdf application is the right temporary expedient - but someone else will need to write the test for that.

(And yes, I should have done it with a fork and a draft pull request, sorry, but I'm time poor right now and that would have taken me a wee bit longer to set up in VSC.)

pyfive/high_level.py
```diff
@@ -217,8 +224,21 @@ def _get_object_by_address(self, obj_addr):
...
+                full_path = f"{base}/{name}" if base else f"/{name}"
+                # check address without instantiating
+                if link_addr == obj_addr:
+                    # return instantiated object
+                    return grp[name]
+                # descend only if it's a subgroup (need to instantiate minimally)
+                # would be neat to have .is_group
+                dataobjs = DataObjects(self.file._fh, link_addr)
+                if not dataobjs.is_dataset and not dataobjs.is_datatype:
+                    queue.append((full_path, grp[name]))
```
kmuehlbauer: This gets the object by address without instantiating completely. The DataObject creation is needed, to check for group and step into groups. Not sure if this more performant, but it doesn't instantiate the complete Datasets.

kmuehlbauer:
JFTR, as enum type is now handled in another PR, I'm concentrating here in the dereferencing of REFERENCE. This was done by using visititems which instantiated all objects which have been iterated over. 

This now just compares the addresses and do a minimal test for group, where I added a respective `is_group`-method.

kmuehlbauer:
Ready for review from my side. Thanks for all the suggestions and details.

kmuehlbauer:
I was able to gain full coverage by adding another test. I'll leave this be now, so reviewers can have a look.

davidhassell:
Thanks, @kmuehlbauer - I'm paying attention again now :), and will have a look at this PR today ...

davidhassell (approved):
Hi @kmuehlbauer,

Thank you for this change. The tests uses cases look good, the tests pass for me too. The new code makes sense, and I always like a corrected spelling mistake :). 

We'll get this merged and out in a new PyPi release (0.5.1) today or tomorrow, so that it can be immediately used in the h5netcdf pyfive backend PR (https://github.com/h5netcdf/h5netcdf/pull/273).

Cheers,
David

valeriupredoi:
indeed, great many thanks @kmuehlbauer and @davidhassell :beer: x2

PR #86, opened 2025-08-20
Fix calculation of padded name length for compound datatype members

bmaranville:
## Description

When reading the names of members of the compound datatype, the calculated name string length has an off-by-one error (doesn't take into account the terminating null character), and so the padded length will incorrectly calculated if the name string (without the null) is a mutiple of 8.

**Note**
This change won't make any difference in the current operation, since the only type of compound dataset that is handled is the special case of complex numbers where the names of the members are 'r' and 'i' for real and imaginary, which won't be affected by this bug.  This bug would only be exposed if more general handling for compound types was implemented in pyfive.

## Before you get started

_I didn't create an issue, because it would say the same thing as above, and it's a 3-character fix_

## Checklist

- [ ] All tests pass

### DIFF
--- a/pyfive/datatype_msg.py
+++ b/pyfive/datatype_msg.py
@@ -112,7 +112,7 @@ class DatatypeMessage(object):
         members = []
         for _ in range(n_comp):
             null_location = self.buf.index(b'\x00', self.offset)
-            name_size = _padded_size(null_location - self.offset, 8)
+            name_size = _padded_size(null_location - self.offset + 1, 8)
             name = self.buf[self.offset:self.offset+name_size]
             name = name.strip(b'\x00').decode('utf-8')
             self.offset += name_size


bmaranville:
We can't really test this change, because the pyfive library rejects any compound member name that is not 'i' or 'r', so we can't independently test a name with a length that is a multiple of 8

We could and should make a test for ENUMERATED names that are multiples of 8 - I can confirm that without adding 1 to `null_location - self.offset` when calculating the name length, you get corrupted data if one of the names has length that is 8, e.g. "GREENISH".

bnlawrence:
We can probably properly test that in the enumeration where we use the same logic. Maybe if we refactor that out into a common function to both bits, that we can test?

bnlawrence:
That being so, I'd be happy for this to be merged now. The only problem is that V is going on holiday in 10 minutes, so there's not much point :-)

Issue #100, opened 2025-09-25
Issue loading file due to buffer too small


valeriupredoi:
Here is the MRE:

```python
import pyfive


def _load_nc_file(ncvar):
    """
    Get the netcdf file and its b-tree.

    Fixture to test loading an issue file.
    """
    issue_file = "da193a_25_6hr_t_pt_cordex__198807-198807.nc"
    nc = pyfive.File(issue_file)
    ds = nc[ncvar]

    return ds


def test_buffer_issue():
    print("File issue: da193a_25_6hr_t_pt_cordex__198807-198807.nc")
    print("Variable m01s30i111")
    _load_nc_file('m01s30i111')

test_buffer_issue()
```

this instantly returns the barf:

```
$ python test_buffer_local.py
File issue: da193a_25_6hr_t_pt_cordex__198807-198807.nc
Variable m01s30i111
Traceback (most recent call last):
  File "/home/valeriu/test_buffer_local.py", line 22, in <module>
    test_buffer_issue()
    ~~~~~~~~~~~~~~~~~^^
  File "/home/valeriu/test_buffer_local.py", line 20, in test_buffer_issue
    _load_nc_file('m01s30i111')
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/home/valeriu/test_buffer_local.py", line 12, in _load_nc_file
    ds = nc[ncvar]
         ~~^^^^^^^
  File "/home/valeriu/miniconda3/envs/pyactive4/lib/python3.13/site-packages/pyfive/high_level.py", line 94, in __getitem__
    return Dataset(obj_name, DatasetID(dataobjs), self)
                             ~~~~~~~~~^^^^^^^^^^
  File "/home/valeriu/miniconda3/envs/pyactive4/lib/python3.13/site-packages/pyfive/h5d.py", line 93, in __init__
    self._meta = DatasetMeta(dataobject)
                 ~~~~~~~~~~~^^^^^^^^^^^^
  File "/home/valeriu/miniconda3/envs/pyactive4/lib/python3.13/site-packages/pyfive/h5d.py", line 593, in __init__
    self.attributes = dataobject.get_attributes()
                      ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/valeriu/miniconda3/envs/pyactive4/lib/python3.13/site-packages/pyfive/dataobjects.py", line 154, in get_attributes
[...]

PR #102, opened 2025-09-28
fix enum attribute

### LINKED ISSUE #101: Error Reading Enum Attributes
```python
import h5py
import pyfive
import numpy as np


f = h5py.File('enum_attrs.hdf5', 'w')
attrs = f.attrs

# Define an enum dtype
enum_dtype = h5py.special_dtype(
    enum=(np.int32, {'one': 1, 'two': 2, 'three': 3})
)
# Create an attribute with that enum dtype
attrs.create('enum', 2, dtype=enum_dtype)

f.close()

with pyfive.File('enum_attrs.hdf5') as hfile:
    import h5py
    enum_attr = hfile.attrs['enum']
    assert enum_attr == 2
    assert enum_attr.dtype == enum_dtype
    )
```

```python
  File "/home/kai/mambaforge/envs/xr_312_np2/lib/python3.12/site-packages/IPython/core/interactiveshell.py", line 3577, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-2-4c07ea7bf2a5>", line 20, in <module>
    enum_attr = hfile.attrs['enum']
                ^^^^^^^^^^^
  File "/home/kai/PycharmProjects/pyfive/pyfive/high_level.py", line 158, in attrs
    self._attrs = self._dataobjects.get_attributes()
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kai/PycharmProjects/pyfive/pyfive/dataobjects.py", line 147, in get_attributes
    name, value = self.unpack_attribute(offset)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kai/PycharmProjects/pyfive/pyfive/dataobjects.py", line 187, in unpack_attribute
    return self._parse_attribute_msg(self.msg_data, offset)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kai/PycharmProjects/pyfive/pyfive/dataobjects.py", line 240, in _parse_attribute_msg
    value = value[0]
            ~~~~~^^^
KeyError: 'There are no fields in dtype int32.'
```

I traced this back to a mismatch in `DataObjects._attr_value`, where instead of the value, the datatype of an enum is returned.

kmuehlbauer:
## Description

This fixes an issue with enum attributes, where instead of the value, the dtype was returned.

Closes #101

While doing this, I had to enhance the tests and according make/data file. Also, I had to fix some inconsistencies in that particular test file wrt numpy.

## Before you get started

## Checklist

- [ ] All tests pass


tests/test_attr_datatypes.py
```diff
@@ -79,10 +79,14 @@ def test_numeric_array_attr_datatypes():
...
-        assert hfile.attrs['vlen_str_array'][1] == b'World!'
 
-        assert hfile.attrs['vlen_str_array'].dtype == np.dtype('S6')
+def test_string_array_attr_datatypes():
+    with pyfive.File(ATTR_DATATYPES_HDF5_FILE) as hfile:
+        assert hfile.attrs['vlen_str_array'][0] == 'Hello'
+        assert hfile.attrs['vlen_str_array'][1] == 'World!'
+
+        assert hfile.attrs['vlen_str_array'].dtype == np.dtype('O')
+        assert hfile.attrs['vlen_str_array'].dtype.metadata == {'h5py_encoding': 'utf-8'}
```
kmuehlbauer: This reads back now as strings. Because I had to fix this, I've taken the chance to move this to its own test function.

tests/make_attr_datatypes_file.py
```diff
@@ -55,20 +55,20 @@
 attrs['vlen_str_array'] = [b'Hello', b'World!']
 
 # variables length sequences
-val = np.empty((2, ), dtype=np.object)
```
kmuehlbauer, line 58: This was complained about in the test suite, so I fixed this as suggested.

pyfive/dataobjects.py
```diff
@@ -250,6 +250,12 @@ def _parse_attribute_msg(self, buffer, offset):
 
     def _attr_value(self, dtype, buf, count, offset):
         """ Retrieve an HDF5 attribute value from a buffer. """
+
+        # first handle ENUMERATION, we just extract the dtype
+        if isinstance(dtype, tuple):
```
kmuehlbauer, line 262: Extracting this in that early stage here works just fine. This then runs the normal path and seems to fix the specific problem for enum attributes.

kmuehlbauer (commented):
I've left some comments for clarification.

pyfive/dataobjects.py
```diff
@@ -230,11 +231,17 @@ def _parse_attribute_msg(self, buffer, offset):
...
         # Read the dataspace information
         shape, maxshape = determine_data_shape(buffer, offset)
-        items = int(np.prod(shape))
-        offset += _padded_size(attr_dict['dataspace_size'], padding_multiple)
 
-        # Read the value(s)
-        value = self._attr_value(dtype, buffer, items, offset)
+        # detect Empty/NULL dataspace
+        if shape is None:
+            value = Empty(dtype=dtype)
```
kmuehlbauer, line 237: After detecting NULL dataspace in `determine_data_shape`, we create Empty value here.

pyfive/dataobjects.py
```diff
@@ -722,6 +733,11 @@ def determine_data_shape(buf, offset):
         offset += DATASPACE_MSG_HEADER_V2_SIZE
     else:
         raise InvalidHDF5File('unknown dataspace message version')
+    # to detect Empty aka NULL dataspace we need to check for V2 DATASPACE MESSAGE
+    if header["version"] == 2:
+        # check for NULL dataspace and return early
+        if header["type"] == 2:
+            return None, None
```
kmuehlbauer: That's the crucial part, see https://support.hdfgroup.org/documentation/hdf5/latest/_f_m_t2.html#subsubsec_fmt2_dataobject_hdr_msg_simple section V2 for details.

kmuehlbauer (commented):
I've added a fix for #100 as it fits thematically in here. Please let me know, if you are K with this.

valeriupredoi:
woohoo @kmuehlbauer you're an absolute legend, was bashing me brains about #100 on Friday, many thanks! Would you mind if I added an integration test with a real (but small) file that replicates that issue? :beers:

valeriupredoi:
> Yes, please go ahead. 🚀
> 
> I'm currently working on ironing out some more issues with EnumType and other user types (eg. COMPOUND).

brill! Should I just PR it to your fork? I still need to get the correct (small) file, I was banking on Iris to chop the original file - which I did, but Iris has fixed that NULL (which is interesting to see it doing that)

kmuehlbauer:
@valeriupredoi Just push any relevant changes directly to my PR. No need to fiddle with PR's here.

valeriupredoi (approved):
this fixes #100 very nicely - I tested both locally (POSIX) and remotely (S3), very many thanks @kmuehlbauer 🍻

kmuehlbauer:
> OK @kmuehlbauer I've been trying to write a file with that NULL DATATSPACE for a wee while now - no success, netCDF4 is actually writing it but adds a point to the data, and if you try pass a Null/None, it complains - do you know how to create one? Got no idea how that file was created to have such a peculiar attribute/null value 🍻

I've added one attribute `empty_string` to `attr_datatypes.hdf5` (and the respective make-file). I'm not sure, when this is triggered. as this attribute is just a global attribute, no attribute of a dataset. Not sure, if that makes a difference. But yes, you could check if this breaks without the fix.

tests/test_buffer_issue.py
```diff
@@ -0,0 +1,38 @@
...
+
+def test_buffer_issue():
+    """
+    Test the case when the attribute contains no data.
+
+    This happens when DATASPACE is NULL and DATA is empty.
+    """
+    print("File with issue da193a_25_6hr_t_pt_cordex__198807-198807.nc")
+    print("Variable m01s30i111")
+    _load_nc_file('m01s30i111')
```
valeriupredoi, line 38: @kmuehlbauer in lieu of an actual small file I've added a test case for an S3-hosted file - this works beautifully with this PR

kmuehlbauer: Nice!

valeriupredoi: if you ready with this PR, I reckon we can merge, bud. We should definitely prepare a new release with this and #105

kmuehlbauer: Yes, I think this is ready for review ;-) For #105 I need to add a bit more testing, but should be ready soon.

valeriupredoi: much obliged, I did review it to the best of my knowledge, I'll ping @bnlawrence below for a review as well :beer:

pyfive/dataobjects.py
bnlawrence: It'd be good to fix the docstring for empty, as we're now actually using it!

pyfive/dataobjects.py
bnlawrence: I'd be tempted to move the `if header['type]] == 2; return None, None` into the version elif ... but I'm also happy to leave it here.

tests/test_attr_datatypes.py
bnlawrence: So, I think this is a different behaviour from h5py. 
 ```
with h5py.File(ATTR_DATATYPES_HDF5_FILE) as hfile:
    ...:     x=hfile.attrs['vlen_str_array'][0]
    ...:     print(x)
b'Hello'
```
so I'm thinking this is not what we want?

**Review by bnlawrence — COMMENTED** [review]
I'm slightly concerned that pyfive will now be doing different things with strings? (The bug fix is fine though!)

File: tests/test_attr_datatypes.py (hunk 1 above)
Comment (kmuehlbauer) [reply tests/test_attr_datatypes.py]: I'll double check this one.

Comment (kmuehlbauer) [reply tests/test_attr_datatypes.py]: @bnlawrence This is obviously new behaviour of h5py introduced some time between 2017 and now. If we recreate the file now, the following line of code:

```python
attrs['vlen_str_array'] = [b'Hello', b'World!']
```

translates to this HDF5 dump:

```
ATTRIBUTE "vlen_str_array" {
   DATATYPE  H5T_STRING {
      STRSIZE H5T_VARIABLE;
      STRPAD H5T_STR_NULLTERM;
      CSET H5T_CSET_ASCII;
      CTYPE H5T_C_S1;
   }
   DATASPACE  SIMPLE { ( 2 ) / ( 2 ) }
   DATA {
   (0): "Hello", "World!"
   }
}
```

whereas the old file translates to:

```
ATTRIBUTE "vlen_str_array" {
   DATATYPE  H5T_STRING {
      STRSIZE 6;
      STRPAD H5T_STR_NULLPAD;
      CSET H5T_CSET_ASCII;
      CTYPE H5T_C_S1;
   }
   DATASPACE  SIMPLE { ( 2 ) / ( 2 ) }
   DATA {
   (0): "Hello\000", "World!"
   }
}

```

So from my perspective this reveals a big flaw in our testing suite. Although we should be able to read ancient files, we should not expect that this gives the same output with current versions of h5py/libhdf5.

I've managed to create both of the above representations with recent h5py and added the two options for now to the tests.

kmuehlbauer: And from looking at this, the old version doesn't seem to be variable length at all.

pyfive/dataobjects.py
kmuehlbauer: Done!

kmuehlbauer (commented):
@bnlawrence I've worked along your suggestions.

pyfive/dataobjects.py
kmuehlbauer: Done

kmuehlbauer:
@bnlawrence According to our discussion in #108 I've reverted the changes of the ingested hdf5 file. Instead I only changed the according make file, taking now an argument destination file path.

For testing I introduced a fixture, which creates that file on demand and tests the necessary added features in new functions.

Is this a way forward? Slight changes to the make scripts and only some changes to the tests.

bnlawrence:
Great. Thanks. @valeriupredoi do your thing.

valeriupredoi:
Very many thanks @kmuehlbauer and @bnlawrence 🍺🍺

PR #105, opened 2025-09-29
Fix user datatypes (enum, compound)

### LINKED ISSUE #104: All commited types are interpreted as EnumTypeID
Currently all committed types are piped through pyfive.h5py.Datatype. There all types are wrapped in `?`EnumTypeID`, which currently breaks for non-enum, like `Compound`.

https://github.com/NCAS-CMS/pyfive/blob/fefabbad69e5eedb41c570bd67cb8ea0ed75cfd8/pyfive/h5py.py#L12-L22

kmuehlbauer:
## Description

This adapts the Type handling to acknowledge `CompoundTypes` (only `complex` for now), `EnumTypes` and generic types as `Datatype`. This fixes several issues of the pyfive implementation over at `h5netcdf`.

Closes #104

Att: This builds on #102 as those preparations are relied on here.

## Before you get started

## Checklist

- [x] More unit tests may be needed
- [x] All tests pass

### DIFF
--- a/pyfive/h5t.py
+++ b/pyfive/h5t.py
@@ -74,7 +74,7 @@ def check_dtype(**kwds):
     else:
         return None
 
-
+# todo: refactor the following classes, so TypeEnumID and TypeCompoundID sublass from the base TypeID.
 class TypeID:
     """
     Used by DataType to expose internal structure of a generic


valeriupredoi:
Very many thanks again @kmuehlbauer :beers: @bnlawrence @davidhassell you folks got some time for a review of this one please?

kmuehlbauer:
Ready for review.

pyfive/dataobjects.py
```diff
@@ -272,8 +289,10 @@ def _attr_value(self, dtype, buf, count, offset):
                     vlen, vlen_data = self._vlen_size_and_data(buf, offset)
                     value[i] = self._attr_value(base_dtype, vlen_data, vlen, 0)
                     offset += 16
-                elif dtype_class == 'ENUMERATION':
-                    return np.dtype(dtype[1],metadata={'enum':dtype[2]})
+                # elif dtype_class == 'ENUMERATION':
+                #     return np.dtype(dtype[1],metadata={'enum':dtype[2]})
+                # elif dtype_class == 'COMPOUND':
+                #     return np.dtype(dtype[1])
```
kmuehlbauer:
```suggestion
```

pyfive/h5t.py
bnlawrence, line 1: I can see that the original implementation of TypeEnumID imagined it was going to be a specialisation of some super class (the super init is still in the init method).  I think the two new classes here look like that really ought to be our strategy. Subclass from a common TypeID? Then quite a lot code duplication would disappear?

bnlawrence:
Thanks very much for this @kmuehlbauer! Great stuff. Apart from my comment on h5t above (which is a matter of taste can could be dealt with in some future refactor), and pending the resolution of the testing issues (#108), I think this is good to go.

pyfive/h5t.py
kmuehlbauer: Yes, I'd tampered a bit with this idea already, but haven't gotten any further. Maybe adding a Todo would be OK for now?

pyfive/h5t.py
```diff
@@ -74,6 +74,36 @@ def check_dtype(**kwds):
     else:
         return None
 
+
```
kmuehlbauer: @bnlawrence Should we add this todo, here, for the time being?

```suggestion
# todo: refactor the following classes, so TypeEnumID and TypeCompoundID sublass from the base TypeID.
```

kmuehlbauer:
I'll do this when #102 is in and I can make a clean rebase/merge. Otherwise I'll go crazy with git.

pyfive/h5t.py
kmuehlbauer: Added a todo comment.

valeriupredoi (approved):
brilliant, great many thanks @kmuehlbauer 🍻

PR #107, opened 2025-09-30
Added partial support for compact datasets.

zequihg50:
## Description

Added partial support for compact datasets (currently NotImplementedError is raised). This supports the data layout messages v3 and v4, v1 and v2 still pending but I can work out this in the near future. GitHub does not allow to attach a real file but I have included tests for a custom compact variable file.

## Checklist

- [X] This pull request has a descriptive title and labels
- [X] This pull request has a minimal description (most was discussed in the issue, but a two-liner description is still desirable)
- [X] Unit tests have been added (if codecov test fails)
- [X] Any changed dependencies have been added or removed correctly (if need be)
- [X] If you are working on the documentation, please ensure the [current build](https://app.readthedocs.org/projects/pyfive/builds/) passes
- [ ] All tests pass


pyfive/h5d.py
```diff
@@ -98,7 +98,21 @@ def __init__(self, dataobject, pseudo_chunking_size_MB=4):
         # throws a flake8 wobbly for Python<3.10; match is Py3.10+ syntax
         match self.layout_class:  # noqa
             case 0:  #compact storage
-                raise NotImplementedError("Compact Storage")
+                self.data = None
```
bnlawrence: I think it would be more consistent with the rest of the code if we pushed the compact handling into a method called from here.

pyfive/h5d.py
```diff
@@ -161,7 +175,18 @@ def get_data(self, args, fillvalue):
         # throws a flake8 wobbly for Python<3.10; match is Py3.10+ syntax
         match self.layout_class:  # noqa
             case 0:  #compact storage
-                raise NotImplementedError("Compact Storage")
+                if self.data is None:
```
bnlawrence: Similarly, this functionality would be better in a method.

bnlawrence (changes requested):
This is great, thanks @zequihg50!  Do you think you could just move your compact support into dedicated private functions? I think that would be cleaner.

kmuehlbauer:
@zequihg50 + :100: for containing within one test file.

Issue #108, opened 2025-10-01
Usage of ancient files for testing breaks with recent h5py/libhdf5


kmuehlbauer:
While working on #102 @bnlawrence and I found a glitch when we recreated test files with recent h5py/libhdf5.

See https://github.com/NCAS-CMS/pyfive/pull/102#discussion_r2393563713 for details.

Citing here for visibility:

If we recreate the file now, the following line of code:

```python
attrs['vlen_str_array'] = [b'Hello', b'World!']
```

translates to this HDF5 dump:

```
ATTRIBUTE "vlen_str_array" {
   DATATYPE  H5T_STRING {
      STRSIZE H5T_VARIABLE;
      STRPAD H5T_STR_NULLTERM;
      CSET H5T_CSET_ASCII;
      CTYPE H5T_C_S1;
   }
   DATASPACE  SIMPLE { ( 2 ) / ( 2 ) }
   DATA {
   (0): "Hello", "World!"
   }
}
```

whereas the old file translates to:

```
ATTRIBUTE "vlen_str_array" {
   DATATYPE  H5T_STRING {
      STRSIZE 6;
      STRPAD H5T_STR_NULLPAD;
      CSET H5T_CSET_ASCII;
      CTYPE H5T_C_S1;
   }
   DATASPACE  SIMPLE { ( 2 ) / ( 2 ) }
   DATA {
   (0): "Hello\000", "World!"
   }
}

```

It might make sense to move from the current behaviour of testing already created files (and ingesting them into GIT) to creating test files only when testing in the test suite. This is what ~up~downstream libraries like xarray and h5netcdf are doing. Thoughts?

kmuehlbauer:
Thanks @bnlawrence for sharing your thoughts on that. 

I believe that anything which was written to an hdf5 file in the past, can be (re)created with recent tools (namely `h5py`). But this we'll have to put to the test. And if this can be guaranteed for all times, is at least questionable. I therefore agree, that there can be good reasons to keep files for testing which have been created with old versions (or are just out there).

As a first step we might try to (re)create all existing files and see what else did change over time. For those possible changes we need to find the exact (re)creation incantations, which then should be fed into the new creation-while-testing behaviour. If we can't work out any (re)creation for certain feature would be an indication of keeping those files around.

bnlawrence:
This is great. I'm thinking we probably need a "for maintiner's" section in our docs?

PR #117, opened 2025-10-08
use pytest temporary dir factory to write some of the test hdf5 files and move all fixed hdf5 sample data files to `tests/data`

### LINKED ISSUE #118: Collect all HDF5 sample datasets into a `tests/sample_data` dir
They are all currently scattered inside the main `tests` dir, which is not optimal 🍺

valeriupredoi:
## Description

~The opaque dataset tests produce three HDF5 sample datasets that are currently git orphaned, I am adding them to version control; it'd be really nice if we collected all these sample datasets in a eg `tests/sample_data` directory, rather than have them float around inside the main `tests` test dir - created #118 for that.~

Using pytest's tmp dir factory to temporarily store those files, and not have them under version control. 

Closes #118 

## Before you get started

## Checklist

- [x] All tests pass

### DIFF
--- a/tests/test_netcd4_classic.py
+++ b/tests/test_netcd4_classic.py
@@ -8,7 +8,7 @@ from numpy.testing import assert_array_equal, assert_almost_equal
 import pyfive
 
 DIRNAME = os.path.dirname(__file__)
-NETCDF4_CLASSIC_FILE = os.path.join(DIRNAME, 'netcdf4_classic.nc')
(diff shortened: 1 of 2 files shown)


kmuehlbauer:
@valeriupredoi As I understood #108, we do not want to git ingest new data files, but instead create them while testing.

kmuehlbauer:
I thought, pytest is cleaning this up. Maybe we should make sure to correctly configure this with pytest.

PR #120, opened 2025-10-09
full feature COMPOUND type

### LINKED ISSUE #119: Issue with COMPOUND including REFERENCE
A major building block of NetCDF4 format is the use of HDF5 dimension scales.

A NetCDF4 `dimension` is characterized in HDF5 as a `DATASET` with an `ATTRIBUTE` of type STRING with name `CLASS`and a `DATASPACE` with `DIMENSION_SCALE` as data content. Furthermore this `DATASET`has an `ATTRIBUTE` `REFERENCE_LIST` which dumps like this:

```
ATTRIBUTE "REFERENCE_LIST" {
  DATATYPE  H5T_COMPOUND {
     H5T_REFERENCE { H5T_STD_REF_OBJECT } "dataset";
     H5T_STD_U32LE "dimension";
  }
  DATASPACE  SIMPLE { ( 1 ) / ( 1 ) }
  DATA {
  (0): {
        DATASET 111195198694608 "/subgroup/y_var",
        0
     }
  }
}
```

For some reason we can work with `REFERENCE` as attributes (dereferencing works), but this special REFERENCE inside COMPOUND doesn't work. I'm trying to create some MCVE but still failing to do so. Over at h5netcdf just one test fails, returning wrong values. But the issue is that pyfive can't correctly return the two values from the above example dump (the `REFERENCE` and the number `0`).

I hope that @bnlawrence can make something out of my ramblings. I'll try to come up with an MCVE.

kmuehlbauer:
- fix complex compound
- fix nested references

## Description

This enables the COMPOUND type to work for all compounds. It fixes some aspects of Complex Compound and it enables proper access of nested REFERENCE type. This is especially needed for HDF5/NetCDF4 dimension scale usage.

Closes #119

## Before you get started

## Checklist

- [ ] All tests pass

### DIFF
--- a/pyfive/dataobjects.py
+++ b/pyfive/dataobjects.py
@@ -220,13 +220,11 @@ class DataObjects(object):
         try:
             dtype = DatatypeMessage(buffer, offset).dtype
         except NotImplementedError:
-            if name == 'REFERENCE_LIST':
-                pass #suppress this one, no one actually cares about these as far as I know
-            else:
-                warnings.warn(
-                    f"Attribute {name} type not implemented, set to None."
-                )
+            warnings.warn(
+                f"Attribute {name} type not implemented, set to None."
+            )
             return name, None
+
         offset += _padded_size(attr_dict['datatype_size'], padding_multiple)
 
         # Read the dataspace information
@@ -258,12 +256,29 @@ class DataObjects(object):
     def _attr_value(self, dtype, buf, count, offset):
         """ Retrieve an HDF5 attribute value from a buffer. """
 
-        # first handle ENUMERATION, we just extract the dtype
+        has_ref = False
         if isinstance(dtype, tuple):
+            # first handle ENUMERATION, we just extract the dtype
             if dtype[0] == "ENUMERATION":
                 dtype = np.dtype(dtype[1], metadata={'enum': dtype[2]})
             elif dtype[0] == "COMPOUND":
-                dtype = np.dtype(dtype[1])
+                # for COMPOUND we have normal dtypes
+                if isinstance(dtype[1], (np.dtype, str)):
+                    dtype = np.dtype(dtype[1])
+                # and we have tuples
+                elif isinstance(dtype[1], tuple):
+                    _dtype = dtype[1]
+                    # check for nested REFERENCE
+                    is_ref = [isinstance(d, tuple) and d[0] == "REFERENCE" for d in _dtype[1]]
+                    has_ref = any(is_ref)
+                    formats = [f"V{d[1]}" if is_ref[i] else d for i, d in enumerate(_dtype[1])]
+                    dtype = np.dtype(
+                        {
+                            'names': _dtype[0],
+                            'formats': formats,
+                            'offsets': _dtype[2],
+                            'itemsize': _dtype[3],
+                        })
 
         if isinstance(dtype, tuple):
             dtype_class = dtype[0]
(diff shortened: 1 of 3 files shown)


pyfive/dataobjects.py
```diff
@@ -220,13 +220,11 @@ def _parse_attribute_msg(self, buffer, offset):
         try:
             dtype = DatatypeMessage(buffer, offset).dtype
         except NotImplementedError:
-            if name == 'REFERENCE_LIST':
```
bnlawrence, line 223: Absolutely my fault. Should have returned None.

pyfive/dataobjects.py
```diff
@@ -258,12 +256,29 @@ def _parse_attribute_msg(self, buffer, offset):
...
+                    is_ref = [isinstance(d, tuple) and d[0] == "REFERENCE" for d in _dtype[1]]
+                    has_ref = any(is_ref)
+                    formats = [f"V{d[1]}" if is_ref[i] else d for i, d in enumerate(_dtype[1])]
+                    dtype = np.dtype(
+                        {
+                            'names': _dtype[0],
+                            'formats': formats,
+                            'offsets': _dtype[2],
+                            'itemsize': _dtype[3],
+                        })
```
bnlawrence, line 281: Are we sure that there can be no other type of tuple appearing in a compound? E.g. could someone put an enumeration in here? I'd be tempted to trap any other type.

pyfive/dataobjects.py
```diff
@@ -258,12 +256,29 @@ def _parse_attribute_msg(self, buffer, offset):
...
+                if isinstance(dtype[1], (np.dtype, str)):
+                    dtype = np.dtype(dtype[1])
+                # and we have tuples
+                elif isinstance(dtype[1], tuple):
+                    _dtype = dtype[1]
+                    # check for nested REFERENCE
+                    is_ref = [isinstance(d, tuple) and d[0] == "REFERENCE" for d in _dtype[1]]
+                    has_ref = any(is_ref)
+                    formats = [f"V{d[1]}" if is_ref[i] else d for i, d in enumerate(_dtype[1])]
+                    dtype = np.dtype(
```
bnlawrence, line 275: I'm not sure exactly what is going on, but is this effectively saying that if you have a reference here, you'll simply say it's something binary that you don't know about? So this isn't dereferencing?

bnlawrence: Does the type get sorted later when being de-referenced?

bmaranville: A Reference just contains an address (pointer) - it doesn't know the type of what it's pointing to.  Note that when you get a "REFERENCE" in a datatype message, you have to read the bit fields to determine if it is an object reference (bits 0-3 == 0) or region reference (bits 0-3 == 1), see https://support.hdfgroup.org/documentation/hdf5/latest/_f_m_t11.html#subsubsec_fmt11_dataobject_hdr_dtmessage

A region reference is handled differently from an object reference, and region reference handling is not implemented in pyfive at this time.

pyfive/dataobjects.py
bmaranville: You are correct - compound can be nested with almost anything (e.g. it can contain other compound types, too!), which is why the datatype lookup is recursive: https://github.com/NCAS-CMS/pyfive/blob/ee6ddb7f7c12f87fd156fdd2b29bbb541c5596f5/pyfive/datatype_msg.py#L141

The handling here doesn't reflect that.

kmuehlbauer: Yes, didn't test that. I'll have a look at h5py/hdf5/netcdf for some examples on that.

pyfive/dataobjects.py
kmuehlbauer: Yes, after reading the data, we transform the void buffer into pyfive.Reference. Maybe we can sort out a better way later.

kmuehlbauer: OK, then I'll need to make this work only for object references. Thanks @bmaranville.

pyfive/dataobjects.py
kmuehlbauer: @bmaranville Thanks for highlighting that. My main task was to fix this for REFERENCE_LIST (dimension scale) compounds.

PR #122, opened 2025-10-13
introduce new H5 types to replace current type-tuples

### LINKED ISSUE #119: Issue with COMPOUND including REFERENCE

kmuehlbauer:
- update type handling
- full COMPOUND support (incl. nested, complex compounds and compounds with references)
- other simplifications and deduplication

## Description

The h5types-as-a-tuple approach didn't work out in #119. I got confused trying to implement for the needs of #119. Thanks @lm41 for the hacking session together over the weekend. It wouldn't have been possible without you :heart:.

So, this adds a comprehensive set of H5Type and derived classes. This makes life much easier and code more readable. Although this adds a bit of overhead compared with the basic tuple system, it should not substantially affect performance.

Closes #119, overrides #120

## Before you get started

## Checklist

- [x] All tests pass

### DIFF
--- a/pyfive/dataobjects.py
+++ b/pyfive/dataobjects.py
@@ -17,9 +17,10 @@ from pyfive.core import UNDEFINED_ADDRESS
 from pyfive.btree import BTreeV1Groups, BTreeV1RawDataChunks
 from pyfive.btree import BTreeV2GroupNames, BTreeV2GroupOrders
 from pyfive.btree import BTreeV2AttrCreationOrder, BTreeV2AttrNames
-from pyfive.btree import GZIP_DEFLATE_FILTER, SHUFFLE_FILTER, FLETCH32_FILTER
+from pyfive.btree import GZIP_DEFLATE_FILTER, SHUFFLE_FILTER, FLETCH32_FILTER, LZF_FILTER
 from pyfive.misc_low_level import Heap, SymbolTable, GlobalHeap, FractalHeap, GLOBAL_HEAP_ID, _decode_array, dtype_replace_refs_with_object
 from pyfive.p5t import P5Type, P5StringType, P5CompoundType, P5VlenStringType, P5ReferenceType, P5SequenceType, P5EnumType
+from pyfive.indexing import OrthogonalIndexer, ZarrArrayStub
 from pyfive.h5py import Empty
 from pyfive.h5d import DatasetID
 
@@ -387,6 +388,8 @@ class DataObjects(object):
             return None
         if GZIP_DEFLATE_FILTER in self._filter_ids:
             return 'gzip'
+        elif LZF_FILTER in self._filter_ids:
+            return 'lzf'
         return None
 
     @property
--- a/pyfive/btree.py
+++ b/pyfive/btree.py
@@ -229,9 +229,18 @@ class BTreeV1RawDataChunks(BTreeV1):
                 cls._verify_fletcher32(chunk_buffer)
                 # strip off 4-byte checksum from end of buffer
                 chunk_buffer = chunk_buffer[:-4]
+            elif filter_id == LZF_FILTER:
+                try:
+                    import lzf
+                except ImportError as e:
+                    raise ModuleNotFoundError(
+                        "LZF codec requires optional package 'python-lzf'."
+                    ) from e
+                uncompressed_len = struct.unpack(">H", chunk_buffer[:2])[0]
+                chunk_buffer = lzf.decompress(chunk_buffer, uncompressed_len)
             else:
                 raise NotImplementedError(
-                    "Filter with id: %i import supported" % (filter_id))
+                    "Filter with id: %i import not supported" % (filter_id))
         return chunk_buffer
 
     @staticmethod
@@ -467,6 +476,7 @@ FLETCH32_FILTER = 3
 SZIP_FILTER = 4
 NBIT_FILTER = 5
 SCALEOFFSET_FILTER = 6
+LZF_FILTER = 32000
 
 
 # Attribute message B-Tree node types
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,6 +58,7 @@ test = [
     "h5py",
     "netCDF4",
     "moto",
+    "python-lzf",
     "s3fs>=2025.9.0",  # v-mismatch with boto3 results in ancient s3fs
 ]
 
--- a/tests/test_filter_pipeline_v2.py
+++ b/tests/test_filter_pipeline_v2.py
@@ -18,12 +18,15 @@ def generate_data():
 
 
 @pytest.mark.parametrize("chunk_size", [None, (10, 10), (20, 20)], ids=lambda x: f"chunk_{x}")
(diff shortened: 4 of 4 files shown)


kmuehlbauer:
@bnlawrence @bmaranville @valeriupredoi and other's I'd like to hear your opinions on the enhancements in this PR.

This PR should tackle the shortcomings of #120 and resolve #119.

The created classes do not have a corresponding h5py class. They are merely a shim between HDF5 and the pyfive outward facing layers. I did this in the first place to be able to see how and where the tuple-types are handled. 

As those classes are no outward facing API and are only internal implementation detail, it could suffice for now. If need be, we can declare with underscore and/or move this away from `pyfive.h5t`.

One possible path of refactoring would be to completely implement these types as pyfive counterpart of `h5py.h5t.TypeID`. This is already done for all types as `pyfive.h5t.TypeID` and special cased for `TypeEnumID` and `TypeCompoundID` (for compatibility with `h5netcdf`). Some of the internal information is only kept via numpy dtype. So no harm to keep the proposed approach and see how it goes.

A quick introduction on the PR:

- Implementation of H5Type, H5IntegerType, H5FloatType, H5String, H5OpaqueType, H5CompoundType , H5ReferenceType, H5EnumType, H5Sequence (`H5T_VLEN`)
- Implementation of derived types and needed subtypes H5FixedStringType, H5VlenStringType, H5CompoundField
- All checks for branching are currently done as `isinstance` checks (we can use `__slots__` to come in the range of the current check of the tuple-type)
- The numpy `dtype` is property of that class and created on accessing that property, so will be lazy until needed (this simplifies the code at several places)
- No dual use of `dtype` (tuple vs array-protocol typestring), but we might think about renaming the `dtype` used for the new types, so that `dtype ` always refer to numpy dtype.
- Datatype message type determination is simplified and returns an instance of a new type or a nested representation (eg. for VLEN/Sequence, Compound, Enum)
- Complex Types are handled as compounds, checks are placed into the new type class, but we still can move this out into an own class
- If storage type and numpy dtype do not fit (eg for references) there is currently a special check which translates the original data into the object type representation. This is especially problematic for nested data. There is room for improvement in that regard.

I've added other alignments and code deduplication as I saw fit.

One remark about the added typing, my IDE automatically adds these kinds of typing and I left it. How should this be tackled? I'm inclined to remove the typing, or make it available for all touched code. What's your opinion on that?

valeriupredoi:
> Trying to test in Python 3.14, too.

Works well 😃

bnlawrence:
I've had a look at this, and like moving away from tuples!

My only concern is one I had even before this enhancement, and one that you've got in your list of things,  and that's that we are overusing the attribute name dtype. 

Effectively we have two types of dtypes and we have to work out which one we are dealing with in any context almost by inspection. The name doesn't help us. That makes reading and understanding the pull request hard, but that hardness is not the fault of the pull request itself, even looking the code requires a lot in our mental stacks. E.g.

https://github.com/NCAS-CMS/pyfive/blob/45e1a9876bfd24f10fdff3283b661630d6d881df/pyfive/dataobjects.py#L255-L259

The comment at least tells us what the local _dtype is for but down at lines 

https://github.com/NCAS-CMS/pyfive/blob/45e1a9876bfd24f10fdff3283b661630d6d881df/pyfive/dataobjects.py#L289-L292

We are not really sure what new_dtype actually is any more.

In most of the pyfive code, dtype is our dtype (was tuple, now an H5T object, which like I say, I like), except like the last snippet where we need the actual type for numpy, and when we expose the dtype to users where again, it's a numpy thing.

Given that your touching every usage of our internal tuple dtype, could we change the name in some way so we don't have this confusion? It feels like this is our best chance to do it and given your list of bullets above, it seems like you do think it, so I reckon it's worth the little refactor to do it now.

(In terms of the IDE adding types, I'd be tempted to remove it for now, if we're going to go there - with linter changes - we can do it all at the same time.)

bnlawrence:
I am inclined not to worry too much about our `h5t` internals differing from those in h5py, but history suggests I am bad at choosing when not to worry ... one way to handle that would be to put all our types in `p5t.py`.

pyfive/dataobjects.py
```diff
@@ -255,42 +252,44 @@ def _parse_attribute_msg(self, buffer, offset):
...
-                dtype = np.dtype(dtype[1])
-
-        if isinstance(dtype, tuple):
-            dtype_class = dtype[0]
-            if dtype_class == 'VLEN_STRING':
-                fdtype = np.dtype('O', metadata={'h5py_encoding': 'utf-8'})
-            else:
-                fdtype=  np.dtype('O')
+        # numpy storage dtype
+        _dtype = ptype.dtype
```
kmuehlbauer: Can now do better here and remove the underscore.

pyfive/h5d.py
```diff
@@ -84,21 +85,7 @@ def __init__(self, dataobject, pseudo_chunking_size_MB=4):
...
-            elif dtype[0] == 'OPAQUE':
-                if dtype[1].startswith('NUMPY:'):
-                    self._dtype = np.dtype(dtype[1][6:], metadata={'h5py_opaque': True})
-                else: 
-                    self._dtype = np.dtype('V'+str(dtype[2]), metadata={'h5py_opaque': True})
-            else:
-                self._dtype = dtype
-        else:
-            self._dtype = np.dtype(dtype)
+        self._ptype = dataobject.ptype
```
kmuehlbauer, line 88: Might also just use without underscore here.

pyfive/dataobjects.py
kmuehlbauer: Done

pyfive/h5d.py
kmuehlbauer: Not sure, everything else is underscored here, too. Leaving it.

kmuehlbauer:
@bnlawrence I've worked along your suggestion on renaming/refactoring. Just from looking at it, it was the right decision.

PR #138, opened 2025-10-27
Optimise when we get access to b-tree by providing lazier view of datasets, access to b-tree location, and new p5dump

### LINKED ISSUE #130: Exposing b-tree layout information
For CMIP7, we need a tool which can tell us whether or not the b-tree is contiguous for a particular variable. This can be trivially done (for any variable) with a very minor modification to pyfive which exposes information that is already being gathered for other purposes.

We propose to add a property method to a the DatasetID which exposes a small dictionary of b-tree statistics:
 ```python
     {'contiguous':Boolean, 'start_address':int,'last_address':int}
 ```

The addresses could be trivially obtained in the existing build_index method which loops over all the addresses, and whether or not they aer contiguous can be calculated by a simple comparison of the range of addresses and the number of entries.

This will be useful for anyone considering whether or not a given HDF5 file will play nicely on object store (or behind a range-get http server).

### LINKED ISSUE #134: A pyfive "ncdump"
One of our objectives with pyfive is to avoid having to have a C-library hanging around where we have our pure-python. In that context it would be nice to include a lightweight equivalent of h5dump or ncdump.

For a developer, h5dump gives more deep information, but in most cases where we have dveelopers, we do have the c-library, so I am tempted to suggest we mimic functionality a bit more like ncdump (though of course we could do both).

Anyway, this proposed enhancement is to deliver something which has the functionality of ncdump, that is, it identifies dimensions and variables and provides key attribute information.

In building a prototype for this functionality I realised we have a problem with being a bit too aggressive in the b-tree caching. More on that in another issue.

### LINKED ISSUE #135: Over agressive optimisation of b-tree reading.
Currently, whenever we do something like

```
f = pyfive.File(file_path) 
 for name in f.keys():
        ds = f[name]
        ... stuff ...
```
we are reading all the metadata and b-tree for each dataset.  This is not optimal for use-cases like #134 (particularly on object store data).

This is fine in many cases, but not all, and in particular, CMIP data (multi GB files) with multiple coordinate variables on object store take an eternity (well, it feels like it), even when chunked nicely.

We made a big deal about the utility of having done this as it means that we can close the file at this point and use the `ds` variable later for read operations (which can open and close the file cheaply if all they are doing is opening/closing at the filesystem level - even on object stores, provided we are within an `s3fs` filesystem context).

It seems like we could and should row back a bit. There are effectively variants of two options I think:

1. We don't load the b-tree until the first instance of reading data (but we don't get the benefit of having closed the file, which we think is helpful in a complex parallel environment).
2. We introduce a new (or modify an) API somewhere to lazily load dataset metadata without instantiating the b-trees.

bnlawrence:
## Description

This pull request addresses three specific issues:
   -  Closes #135:  what used to happen was that the b-tree had to be loaded as soon as you wanted to inspect any dataset attributes. This is still the default behaviour, but now there is an additional option to get a lazy view of the variable which doesn't load the b-tree, this lazy support is propagated into visititems as well.
   - Closes #130: the issue here was that we want to make information about the b-tree location available for inspection (to help diagnose performance problems when accessing files). This is a simple fix (new attribute on DatsetID).
   - Closes #134: Introducing a p5dump so that we can inspect a file without any c-code (and potentially inspect many files in parallel).

@zequihg50 Could you please have a look at this one too? Especially the b-tree range stuff and the related tests.

## Checklist

- [x] All tests pass

### DIFF
--- a/doc/p5dump.rst
+++ b/doc/p5dump.rst
@@ -11,7 +11,7 @@ When called with `-s` (e.g ``p5dump -s myfile.hdf5``, or simply ``p5dump myfile.
 datasets, including the locations of the start and end of the chunk index b-tree 
 and the location of the first data chunk for that variable. This extra information is useful for understanding
 the performance of data access for chunked variables, particularly when accessing data in object stores such as
-S3. In general, if one fineds that the b-tree index continues past the first data chunk, access 
+S3. In general, if one finds that the b-tree index continues past the first data chunk, access 
 performance may be sub-optimal - in this situation, if you have control over the data, you might well
 consider using the ``h5repack`` tool from the standard HDF5 distribution to make a copy of the file with the 
 chunk index and attributes stored contiguously.  All tools which read HDF5 files will benefit from this.


.vscode/settings.json
kmuehlbauer, line 1: That slipped in by accident?

pyfive/h5d.py
```diff
@@ -218,18 +247,33 @@ def convert_slice(aslice):
...
+        if self.layout_class != 2:
+            raise TypeError("Data is not chunked") 
+        if not self._index:
+            self._build_index()
+        return self._index
+        
+
+    ##### This property is made available to help understand object store performance
+    @property
+    def btree_range(self):
```
davidhassell: Could we also have `def first_chunk(self):` and `def n_chunks(self):` properties?

pyfive/inspect.py
```diff
@@ -0,0 +1,201 @@
...
+        for g,o in groups.items():
+            safe_print(f'{indent}group: {g} '+'{')
+            gindent = indent+' '
+            dump_header(o,gindent,real_dimensions, special=special)
+            safe_print(gindent+'}'+f' // group {g}')
+   
+
+
+def p5ncdump(file_path, special=False):
+
```
davidhassell, line 192: Some sort of docstring ....

pyfive/h5d.py
```diff
@@ -218,18 +247,33 @@ def convert_slice(aslice):
...
+        """ A tuple with the addresses of the first b-tree node
+        for this variable, and the address of the furthest away node
+        (Which may not be the last one in the chunk index). This property
+        may be of use in understanding the read performance of chunked
+        data in object stores.  ``btree_range`` is a ``pyfive`` API extgension.
+        """
+        self.__chunk_init_check()
+        return (self._btree_start, self._btree_end)
+
```
davidhassell, line 275:
```suggestion

    ##### This property is made available to help understand object store performance
    @property
    def first_chunk(self):
        """The integer address of the first data chunk for this variable.
        
        This property may be of use in understanding the read
        performance of chunked data in object stores.  ``first_chunk``
        is a ``pyfive`` API extension.

        """
        self.__chunk_init_check()
        return self.get_chunk_info(0).byte_offset
```

pyfive/h5d.py
```diff
@@ -218,18 +247,33 @@ def convert_slice(aslice):
...
+        
+
+    ##### This property is made available to help understand object store performance
+    @property
+    def btree_range(self):
+        """ A tuple with the addresses of the first b-tree node
+        for this variable, and the address of the furthest away node
+        (Which may not be the last one in the chunk index). This property
+        may be of use in understanding the read performance of chunked
+        data in object stores.  ``btree_range`` is a ``pyfive`` API extgension.
```
davidhassell:
```suggestion
        data in object stores.  ``btree_range`` is a ``pyfive`` API extension.
```

pyfive/inspect.py
```diff
@@ -0,0 +1,201 @@
...
+        printattr(ds.attrs, ommit)
+
+        if special:
+            extras = {'_Storage':{0:'Compact',1:'Contiguous',2:'Chunked'}[ds.id.layout_class]}
+            if ds.id.layout_class==2:
+                extras['_n_chunks'] = ds.id.get_num_chunks()
+                if extras['_n_chunks'] != 0:
+                    extras['_chunk_shape'] = ds.id.chunks
+                    extras['_btree_range'] = ds.id.btree_range
+                    extras['_first_chunk'] = ds.id.get_chunk_info(0).byte_offset
```
davidhassell: Given the new `first_chunk` property I suggested in `h5d.py`:
```suggestion
                    extras['_first_chunk'] = ds.id.first_chunk
```

tests/test_chunk_index_options.py
```diff
@@ -0,0 +1,122 @@
...
+            ds.id.get_chunk_info(0)
+
+        with pytest.raises(TypeError):
+            ds.id.get_num_chunks()
+
+        with pytest.raises(TypeError):
+            ds.id.read_direct_chunk(0)
+
+        with pytest.raises(TypeError):
+            ds.id.btree_range
```
davidhassell, line 107:
```suggestion
            ds.id.btree_range

        with pytest.raises(TypeError):
            ds.id.first_chunk

```

tests/test_chunk_index_options.py
```diff
@@ -0,0 +1,122 @@
...
+        assert ds.id._DatasetID__index_built==False
+
+        si = StoreInfo((0,0), 0, 4016, 16)
+        info = ds.id.get_chunk_info(0)
+        assert info == si
+        assert h5f["dataset1"].id.get_chunk_info(0) == si
+
+        assert ds.id.get_num_chunks() == 88
+        assert h5f["dataset1"].id.get_num_chunks() == 88
+
```
davidhassell, line 80:
```suggestion
    
        assert ds.id.first_chunk == si.byte_offset

```

pyfive/p5dump.py
```diff
@@ -0,0 +1,49 @@
+from pyfive import p5ncdump
+import sys
+import signal
+
+def main(argv=None):
+    """
+    Provides some of the functionality of tools like ncdump and h5dump.
+    By default this will attempt to do something similar to ncdump.
+    - h will return this information
+    - will provide additional information
```
davidhassell:
```suggestion
    -s will provide additional information
```

kmuehlbauer (commented):
Can't say much about the changes to btree, but the p5dump works smoothly on my system.

bnlawrence:
@valeriupredoi I think i've made all the requested changes. Can you please also have a look at the docs and make sure you're happy with the changes? Otherwise I think this is good to go.

tests/test_mock_s3fs.py
```diff
@@ -57,3 +57,12 @@ def test_s3file_with_s3fs(s3fs_s3):
         pyfive_ds = pyfive.File(f)
         print(f"Dataset loaded from mock S3 with s3fs and Pyfive: ds")
         assert "q" in pyfive_ds
+
+    # test command command line main test_s3
+    with s3.open(os.path.join("MY_BUCKET", file_name), "rb") as f:
+        pyfive.p5ncdump(f)
```
valeriupredoi, line 63: this would be nice to be run not as a call to a Python module, but rather, as a call to a command line executable, I can show you how to do that within a pytest test

bnlawrence: I suspect we should have *both* as tests! Please just provide a suggested addition.

valeriupredoi: Absolutely! I was too frugal with my words yesterday, will pop it in today when I do the full review (in a wee bit)

doc/p5dump.rst
```diff
@@ -0,0 +1,19 @@
+p5dump
+******
+
+``pyfive`` includes a command line tool ``p5dump`` which can be used to dump the contents of an HDF5 file to the 
+terminal. This is similer to the ``ncdump`` tool included with the NetCDF library, or the ``h5dump`` tool included
```
valeriupredoi:
```suggestion
terminal. This is similar to the ``ncdump`` tool included with the NetCDF library, or the ``h5dump`` tool included 
```

doc/p5dump.rst
```diff
@@ -0,0 +1,19 @@
+p5dump
+******
+
+``pyfive`` includes a command line tool ``p5dump`` which can be used to dump the contents of an HDF5 file to the 
+terminal. This is similer to the ``ncdump`` tool included with the NetCDF library, or the ``h5dump`` tool included 
+with the HDF5 library, but like the rest of pyfive, is implemented in pure Python without any dependencies on the 
+HDF5 C library.
+
+It is not identical to either of these toosl, though the default output is very close to that of ``ncdump``.
```
valeriupredoi:
```suggestion
It is not identical to either of these tools, though the default output is very close to that of ``ncdump``.
```

doc/p5dump.rst
```diff
@@ -0,0 +1,19 @@
...
+When called with `-s` (e.g ``p5dump -s myfile.h5``) the output provides extra information for chunked
```
valeriupredoi:
```suggestion
When called with `-s` (e.g ``p5dump -s myfile.hdf5``, or simply ``p5dump myfile.hdf5``) the output provides extra information for chunked
```

doc/p5dump.rst
```diff
@@ -0,0 +1,19 @@
...
+datsets, including the locations of the start and end of the chunk index b-tree
```
valeriupredoi:
```suggestion
datasets, including the locations of the start and end of the chunk index b-tree 
```

doc/p5dump.rst
```diff
@@ -0,0 +1,19 @@
...
+terminal. This is similer to the ``ncdump`` tool included with the NetCDF library, or the ``h5dump`` tool included 
+with the HDF5 library, but like the rest of pyfive, is implemented in pure Python without any dependencies on the 
+HDF5 C library.
+
+It is not identical to either of these toosl, though the default output is very close to that of ``ncdump``.
+When called with `-s` (e.g ``p5dump -s myfile.h5``) the output provides extra information for chunked
+datsets, including the locations of the start and end of the chunk index b-tree 
+and the location of the first data chunk for that variable. This extra information is useful for understanding
+the performance of data access for chunked variables, particularly when accessing data in object stores such as
+S3. In general, if one fineds that the b-tree index continues past the first data chunk, access
```
valeriupredoi:
```suggestion
S3. In general, if one finds that the b-tree index continues past the first data chunk, access 
```

valeriupredoi (commented):
spelling! Brayn 😁

Issue #140, opened 2025-10-27
Flaky test `tests/test_filter_pipeline_v2.py::test_hdf5_filters[fletcher32_True-shuffle_True-compression_9-chunk_(10, 10)]`


valeriupredoi:
Second time we see this, in a fairly large number of trials (30 odd) but there is definitely something flaky in it: https://github.com/NCAS-CMS/pyfive/actions/runs/18843709458/job/53762175768

@kmuehlbauer would you have a minute look into it please, mate? 🍺

PR #142, opened 2025-10-28
add pytest reruns plugin ` pytest-rerunfailures` and minimal settings

valeriupredoi:
## Description

Contributes to avoiding #140 and makes #141 work, as @kmuehlbauer astutely pointed out

## Before you get started

## Checklist

- [x] All tests pass

### DIFF
--- a/tests/test_filter_pipeline_v2.py
+++ b/tests/test_filter_pipeline_v2.py
@@ -17,7 +17,7 @@ def generate_data():
     return np.random.rand(100, 100)
 
 
-@pytest.mark.flaky
+@pytest.mark.flaky(reruns=3, only_rerun="ValueError")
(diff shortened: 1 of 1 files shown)


kmuehlbauer:
OK, so the current setup with `--reruns=3` will rerun any test case which fails, marked or not.

According to my observation, this will only rerun tests once if tests are marked with `@pytest.mark.flaky`.

So we could remove `--reruns=3` to directly catch flaky test and mark each flaky tests like `@pytest.mark.flaky(reruns=3, only_rerun="ValueError")`. 

Update: Because `ValueError` is the error which is raised in the flaky case.

kmuehlbauer (approved):
Let's see how this works!

valeriupredoi:
cheers muchly, Kai! :beer:

PR #145, opened 2025-11-04
Proposal for is_cloud_optimized

zequihg50:
The new API introduced in #138 allows for a mostly trivial implementation of `is_cloud_optimized`. This could serve to validate files that have been [repacked](https://github.com/davidhassell/cmip7repack/). This is just a proposal, if this looks correct I can continue providing a test for it. We should maybe take into account other factors such as not chunked variables.

### DIFF
--- a/pyfive/high_level.py
+++ b/pyfive/high_level.py
@@ -263,7 +263,7 @@ class File(Group):
     @property
     def consolidated_metadata(self):
         """Returns True if all B-tree nodes for chunked datasets are located before the first chunk in the file."""
-        is_co = True
+        is_consolidated = True
         f = self
 
         # for all chunked datasets, check if all btree nodes are located before any dataset chunk
@@ -277,9 +277,9 @@ class File(Group):
                         min_chunk = f[ds].id.first_chunk
 
         if max_btree is not None and min_chunk is not None:
-            is_co = max_btree < min_chunk
+            is_consolidated = max_btree < min_chunk
 
-        return is_co
+        return is_consolidated
 
     def __repr__(self):
         return '<HDF5 file "%s" (mode r)>' % (os.path.basename(self.filename))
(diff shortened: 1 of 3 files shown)


kmuehlbauer:
Thanks @zequihg50 for the details, much appreciated.

pyfive/high_level.py
```diff
@@ -263,6 +263,26 @@ def __init__(self, filename, mode='r'):
         self.userblock_size = 0
         super(File, self).__init__('/', dataobjects, self)
 
+    @property
+    def is_cloud_optimized(self):
```
davidhassell: Need a doc string, that defines what is being tested.

davidhassell: How about `consolidated_internal_file_metadata`
```python
>>> hfile.consolidated_internal_file_metadata
True
```
The doc string can define that "consolidated internal file metadata" that all of the internal file metadata fragments are earlier in the files than the first of all the data chunks.

This is quite a lengthy name, but it is a niche property, and "consolidated internal file metadata" is the phrase that folk at HDF5 use (I got told off for using the term B-tree :-)), so maybe it's better to be strictly correct and descriptive in this case, rather than subjective on the _use_ of the information.

davidhassell: (see below)

zequihg50: `Consolidated metadata` is the term, thanks @davidhassell . It is also what zarr uses to describe a similar problem, so all people should be conceptually familiar with this https://github.com/zarr-developers/zarr-specs/pull/309. To make the name shorter, I would say `def consolidated_metadata` does the job, do you agree?

davidhassell: That's all good!

zequihg50: I have added a small docstring, please feel free to edit if necessary.

valeriupredoi:
thanks @zequihg50 and @davidhassell :beers:

PR #148, opened 2025-11-10
Milestone 1p0

### LINKED ISSUE #137: current versions of pyfive (0.5-0.9) assume a v1 btree for chunked data and never check

bnlawrence:
This is the clean up pull request for the final things for pyfive 1.0
At this point there are only two outstanding issues: #140 and #137. 
When this is complete, both should be closed.
For now, this closes #137.
@valeriupredoi: can you please sort out #140 or make a decision to move it to the Milestone 2.0 collection.

## Checklist

- [ ] All tests pass

### DIFF
--- a/pyfive/dataobjects.py
+++ b/pyfive/dataobjects.py
@@ -527,9 +527,8 @@ class DataObjects(object):
             dims, address = struct.unpack_from(
                 '<BQ', self.msg_data, property_offset)
             data_offset = property_offset + struct.calcsize('<BQ')
-        try:
-            assert (version >= 1) and (version <= 3)
-        except AssertionError:
+        supported_version = (version >= 1) and (version <= 3)
+        if not supported_version:
             # Note that implementation of layout class version 4 will need to deal with
             # assumption that btree is of type V1 (see chunk_btree = ... in h5d.py).
             raise RuntimeError(f'Pyfive cannot yet read HDF5 files with layout class {version}')


pyfive/dataobjects.py
```diff
@@ -527,7 +527,12 @@ def _get_chunk_params(self):
             dims, address = struct.unpack_from(
                 '<BQ', self.msg_data, property_offset)
             data_offset = property_offset + struct.calcsize('<BQ')
-        assert (version >= 1) and (version <= 3)
+        try:
+            assert (version >= 1) and (version <= 3)
+        except AssertionError:
```
valeriupredoi:
```suggestion
        unsupported_ver = (version >= 1) and (version <= 3)
        if unsupported_ver:
```

valeriupredoi: could even go straight to eg `if (version >= 1) and (version <= 3):` to hack off one more line of code, but I like declaring variables for further (possible) conditions

bnlawrence: Not sure that's the correct logic, but I agree it's better than the assertion. Will commit a revision.

valeriupredoi:
@kmuehlbauer we caught that test from #140  fail, but this time around with a `TypeError` https://github.com/NCAS-CMS/pyfive/actions/runs/19300769569/job/55194338618?pr=148 hence it failed-failed, with no reruns

PR #154, opened 2025-11-18
One read for b-tree nodes

### LINKED ISSUE #153: b-tree read performance still not good enough
For stupidly large files we're still seeing slow b-tree reads.

On inspection, this occurs, in part, because we do a lot of very tiny reads which we can avoid by reading entire b-tree nodes in one go.

The offending piece of code is this:

```python
 def _read_node(self, offset, node_level):
        """ Return a single node in the b-tree located at a give offset. """
        node = self._read_node_header(offset, node_level)
        keys = []
        addresses = []
        for _ in range(node['entries_used']):
            chunk_size, filter_mask = struct.unpack('<II', self.fh.read(8))
            fmt = '<' + 'Q' * self.dims
            fmt_size = struct.calcsize(fmt)
            chunk_offset = struct.unpack(fmt, self.fh.read(fmt_size))
            chunk_address = struct.unpack('<Q', self.fh.read(8))[0]
```
There is an obvious optimisation for V1 btrees where we know the length of the entries.

bnlawrence:
## Description

This addresses b-tree read performance issues outlined in #153 by minimising the number of actual reads necessary to read a b-tree node. On some file systems this should make a big difference.

Closes #153

## Checklist

- [x] All tests pass

### DIFF
--- a/pyfive/btree.py
+++ b/pyfive/btree.py
@@ -134,14 +134,29 @@ class BTreeV1RawDataChunks(BTreeV1):
     def _read_node(self, offset, node_level):
         """ Return a single node in the b-tree located at a give offset. """
         node = self._read_node_header(offset, node_level)
+        
         keys = []
         addresses = []
-        for _ in range(node['entries_used']):
-            chunk_size, filter_mask = struct.unpack('<II', self.fh.read(8))
-            fmt = '<' + 'Q' * self.dims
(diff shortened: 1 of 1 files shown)


bnlawrence:
This has yet to be tested on real data and evaluated, but shared as a draft for evaluation and discussion.

PR #160, opened 2025-12-10
Test case for corner case file (buffer too small)

### LINKED ISSUE #158: Possible `pyfive` bug/edge case
### This issue is to cross reference this one from `cmip7repack` https://github.com/NCAS-CMS/cmip7_repack/issues/21

I have been using `check_cmip7_packing` successfully for many files and it has been working great. However, on one particular file it fails with a `CAN'T PARSE` exit code of `5`. 

I have checked this file via a few different methods and it appears to be normal, and have successfully run the check on other similar files that have worked fine. 

After having a dig the problem appears to occur in `pyfive`. Possibly some sort of edge case?

```
>>> import pyfive
>>> f = pyfive.File("filename")
>>> f.consolidated_metadata
  File "<stdin>", line 1, in <module>
  File "<redacted>/lib/python3.10/site-packages/pyfive/high_level.py", line 272, in consolidated_metadata
    if isinstance(f[ds], Dataset):
  File "<redacted>/lib/python3.10/site-packages/pyfive/high_level.py", line 64, in __getitem__
    return self.__getitem_lazy_control(y, noindex=False)
  File "<redacted>/lib/python3.10/site-packages/pyfive/high_level.py", line 129, in __getitem_lazy_control
    return Dataset(obj_name, DatasetID(dataobjs, noindex=noindex), self)
  File "<redacted>/lib/python3.10/site-packages/pyfive/h5d.py", line 96, in __init__
    self._meta = DatasetMeta(dataobject)
  File "<redacted>/lib/python3.10/site-packages/pyfive/h5d.py", line 699, in __init__
    self.attributes = dataobject.get_attributes()
  File "<redacted>/lib/python3.10/site-packages/pyfive/dataobjects.py", line 155, in get_attributes
    more_attrs = self._get_attributes_from_attr_info(attrs, attr_info)
  File "<redacted>/lib/python3.10/site-packages/pyfive/dataobjects.py", line 181, in _get_attributes_from_attr_info
    name, value = self._parse_attribute_msg(data,0)
  File "<redacted>/lib/python3.10/site-packages/pyfive/dataobjects.py", line 197, in _parse_attribute_msg
    version = struct.unpack_from('<B', buffer, offset)[0]
struct.error: unpack_from requires a buffer of at least 1 bytes for unpacking 1 bytes at offset 0 (actual buffer size is 0)
```
We have hosted the file on JASMIN, if you would be able to investigate this it would  be much appreciated:
` ~cdds/tmp/pyfive_error/`

Thankyou!

valeriupredoi:
## Description

@mo-gill raised a similar issue as we have seen and fixed in #100 but with a slightly different type of buffer-related issue, and file. I added the offending file in a test case here so we can have full control over the used case. I suspect it's yet again a NULL vector that's causing it, but I've not looked closely.

Closes #158 

## Before you get started

## Checklist

- [x] All tests pass

### DIFF
--- a/pyfive/misc_low_level.py
+++ b/pyfive/misc_low_level.py
@@ -186,7 +186,7 @@ class FractalHeap(object):
     (i.e. inside a direct block, possibly indexed by one or more indirect blocks). 
 
     Which direct and indirect blocks contains the data, and the offset within the direct 
-    block can be calculated ly using the various parameters and algorithms described 
+    block can be calculated by using the various parameters and algorithms described 
     at the start of the fractal heap section. It is an array of blocks of increasing size 
     within a linear address space.
 


pyfive/misc_low_level.py
```diff
@@ -229,23 +229,48 @@ def __init__(self, fh, offset):
...
         fh.seek(offset)
         data = fh.read(block_size)
         header = _unpack_struct_from(self.direct_block_header, data)
-        header["signature"] == b"FHDB"
-        return data
+        assert header["signature"] == b"FHDB"
+        return data, int.from_bytes(header["block_offset"],
+                                                byteorder="little", signed=False)
+
+    def heapid_to_buffer_offset(self, heapid_offset):
```
kmuehlbauer:
```suggestion
    def _heapid_to_buffer_offset(self, heapid_offset):
```

pyfive/misc_low_level.py
```diff
@@ -262,7 +287,14 @@ def get_data(self, heapid):
                 data_offset += nbytes
                 nbytes = self._managed_object_length_size
                 size = _unpack_integer(nbytes, heapid, data_offset)
-                return self.managed[offset:offset+size]
+
+                # map heap_id offset to flat buffer offset
+                offset = self.heapid_to_buffer_offset(offset)
```
kmuehlbauer:
```suggestion
                offset = self._heapid_to_buffer_offset(offset)
```

kmuehlbauer (changes requested):

bnlawrence:
@kmuehlbauer Thanks for jumping in on this, it was on my to-do for today.  If I've understood what you've suggested ehre, you've dealt with two issues, one is the nothing there, which you've pretty much dealt with by using continue instead of break, and the other is the bookkeeping around not properly getting all the heap information. Does our test case cover both, or have I got that wrong?

kmuehlbauer:
@bnlawrence I'm pretty much into this now. I've discovered another "flaw" when checking for direct/indirect FHDB/FHIB nesting. I'll try to get this fixed, too and add a dedicated test.

pyfive/misc_low_level.py
```diff
@@ -330,7 +366,9 @@ def _indirect_info(self, nrows):
         table_width = self.header['table_width']
         nobjects = nrows * table_width
         ndirect_max = self._max_direct_nrows * table_width
-        if nrows <= ndirect_max:
+        # this info cannot tell the precise amount of blocks
+        # it can just tell us the maximum possible amount we should parse
+        if nobjects <= ndirect_max:
```
kmuehlbauer, line 431: Third "flaw":
This prevented decoding, so we need to take table_width into account.

tests/test_fractal_heap.py
```diff
@@ -0,0 +1,84 @@
...
+import pyfive
+
+@pytest.fixture(scope='module')
+def name(tmp_path_factory):
+    return tmp_path_factory.mktemp("temp") / "fractal_heap.hdf5"
+
+
+@pytest.mark.parametrize("payload_size", [4033, 4032])
+@pytest.mark.parametrize("n_attrs", [10, 11])
+def test_huge_object(name, payload_size, n_attrs):
```
kmuehlbauer, line 14: This is something a stumbled over while trying to generate a test. So for my linux these figures work, but we might need to generalize better.

tests/test_fractal_heap.py
```diff
@@ -0,0 +1,84 @@
...
+
+    with pyfive.File(name, "r") as f:
+        with err:
+            attrs2 = f.attrs
+            print(attrs2.keys())
+
+            for k, v in attrs.items():
+                np.testing.assert_equal(v, attrs2[k])
+
+@pytest.mark.parametrize("n_attrs", [115, 116])
```
kmuehlbauer, line 39: If you debug iterating fractal heap, you will see, that 115 will not have another `FHIB` whereas 116 there is another `FHIB` with one nested `FHDB`. Seems there is no way of checking other than iterating again.

pyfive/misc_low_level.py
```diff
@@ -288,34 +320,38 @@ def _read_integral(self, fh, nbytes):
...
+        # todo: this isn't really clear how the number of ndirect is deduced
+        # at least, we need to derive the correct number by iterating over below
         ndirect, nindirect = self._indirect_info(nrows)
 
         direct_blocks = list()
         for i in range(ndirect):
             address = struct.unpack('<Q', fh.read(8))[0]
             if address == UNDEFINED_ADDRESS:
-                break
+                # if there is no valid address, we move on to the next
```
kmuehlbauer, line 393: So, this was the first issue: There are cases, where attributes are in higher rows, even if lower rows are not set. So iteration over all possible addresses is needed!

pyfive/misc_low_level.py
```diff
@@ -288,34 +320,38 @@ def _read_integral(self, fh, nbytes):
...
+                continue
             block_size = self._calc_block_size(i)
             direct_blocks.append((address, block_size))
 
         indirect_blocks = list()
         for i in range(ndirect, ndirect+nindirect):
             address = struct.unpack('<Q', fh.read(8))[0]
             if address == UNDEFINED_ADDRESS:
-                break
+                # same here, move on to the next address
```
kmuehlbauer, line 402: Same here.

pyfive/misc_low_level.py
```diff
@@ -262,7 +287,14 @@ def get_data(self, heapid):
                 data_offset += nbytes
                 nbytes = self._managed_object_length_size
                 size = _unpack_integer(nbytes, heapid, data_offset)
-                return self.managed[offset:offset+size]
+
+                # map heap_id offset to flat buffer offset
+                offset = self._heapid_to_buffer_offset(offset)
```
kmuehlbauer, line 352: So, this is the second issue: The original code extracts offset and size into the heap, not into out flat managed buffer. When we have attributes in higher rows, this would break. We somehow have to get the correct addresses, we do it by storing them on the heap object.

pyfive/misc_low_level.py
kmuehlbauer: Note: this was only a problem for many, many, many attributes and/or very, very large attributes

valeriupredoi:
brilliant, great many thanks @kmuehlbauer and @bnlawrence :beers: Kai, could I please get an Approve from you on this? Then I could merge :saluting_face:

pyfive/misc_low_level.py
```diff
@@ -168,11 +170,61 @@ def objects(self):
...
+    carerful use of the HDF5 file creation properties.
+    
+    The fractal heap ID can refer to a “tiny”, “huge”, or “managed” object. 
+    If it's tiny, the ID contains the actual data and the heap itself does not need to be read from. 
+    If it's huge, the ID contains the address on disk of the data or a b-tree key that can be used to find this address. 
+    If it's managed, then it contains the offset and length within the virtual fractal heap address space 
+    (i.e. inside a direct block, possibly indexed by one or more indirect blocks). 
+
+    Which direct and indirect blocks contains the data, and the offset within the direct 
+    block can be calculated ly using the various parameters and algorithms described
```
kmuehlbauer:
```suggestion
    block can be calculated by using the various parameters and algorithms described 
```

kmuehlbauer (changes requested):

PR #170, opened 2026-01-05
Allow negative slices when indexing chunked data

### LINKED ISSUE #169: Allow negative slices when indexing chunked data
Currently (v1.0.1) negative slices, i.e. slices with a negative step,  such as `[10:1:-1]`, are not allowed on chunked data. This breaks some downstream applications (e.g. `cfdm` and `cf-python`):

```python
>>> import pyfive
>>> q = pyfive.File('pyfive/tests/data/chunked.hdf5')
>>> q['dataset1']
<HDF5 dataset "dataset1": shape (21, 16), type "int32">
>>> q['dataset1'][::-1]
NegativeStepError: only slices with step >= 1 are supported
```

This can be fixed :) 

Negative slices are already OK for contiguous data (because they uses `np.memmap` access in this case):

```python
>>> import pyfive
>>> p = pyfive.File('pyfive/tests/data/dataset_multidim.hdf5')
>>> p['d']
<HDF5 dataset "d": shape (2, 3, 4, 5), type "int32">
>>> p['d'][::-1, 0, 0, 0]
array([60,  0], dtype=int32)
```

davidhassell:
## Description

Currently (v1.0.1) negative slices, i.e. slices with a negative step, such as `[10:1:-1]`, are not allowed on chunked data. This breaks some downstream applications (e.g. cfdm and cf-python). This can be fixed.

Closes #169 

## Before you get started

## Checklist

- [x] All tests pass


pyfive/indexing.py
```diff
@@ -47,10 +47,6 @@ class VindexInvalidSelectionError(_BaseZarrIndexError):
 class BoundsCheckError(_BaseZarrIndexError):
     _msg = "index out of bounds for dimension with length {0}"
 
-class NegativeStepError(IndexError):
-    def __init__(self):
-        super().__init__("only slices with step >= 1 are supported")
-
 # And the rest of the code is the original file.
```
bnlawrence: I guess this is no longer true

pyfive/indexing.py
```diff
@@ -1018,6 +1047,102 @@ def make_slice_selection(selection):
...
+            dim -= 1
+
+        elif isinstance(index, np.ndarray):
+           if not index.ndim:
+               dim -= 1
+
+        elif index == Ellipsis:
+            raise ValueError(
+                "replace_negative_slices doesn't work when selection "
+                "contains Ellipsis. Consider running replace_ellipsis first"
```
bnlawrence: What is "replace_ellipsis", would the user know about this, could it have a link?

pyfive/indexing.py
```diff
@@ -1018,6 +1047,102 @@ def make_slice_selection(selection):
...
+
+    # Initialize outputs
+    parsed_selection = []
+    reverse_dims = []
+
+    # Dimension positions of the array *after* indexing (some
+    # dimensions might get dropped by integer indices)
+    dim = 0
+
+    # Parse the indices
```
bnlawrence: Maybe it's not so much parse the indices, as "parse and replace the indices"?

bnlawrence (commented):
This looks pretty straightforward I think, but I just have some minor comment comments to consider.

pyfive/indexing.py
davidhassell: Addressed: 082b00fcc1c55ac6913e53c6c545833ccbf74e4f

pyfive/indexing.py
davidhassell: I've changed the message to 
```python
                "replace_negative_slices(selection, shape) doesn't work "
                "when selection contains Ellipsis. Consider doing "
                "selection=pyfive.indexing.replace_ellipsis(selection, shape) "
                "first"
```
bacfb99c45144a8af510463cf07c37157066f550

pyfive/indexing.py
davidhassell: That's right. I think that negative step slices are OK in all cases, now ...

bnlawrence: So I would remove that comment (line 50), then I think this can be approved.

bnlawrence (commented):
I'm good to go once line 50 is gone (I would have done it myself .. .but)

valeriupredoi (approved):

PR #182, opened 2026-01-23
Add __orthogonal_indexing__ flag

### LINKED ISSUE #171: Add `__orthogonal_indexing__` flag
Some libraries include an `__orthogonal_indexing__` property to indicate to the user whether or not a variable utilises orthogonal indexing. For instance, `netCDF4`, `cfdm`, and `cf-python` implements this API, and it's very useful because it tells downstream applications whether or not they themselves need to implement orthogonal indexing.

For these other libraries just mentioned, `__orthogonal_indexing__` is always `True`, but for a pyfive `Dataset`, orthogonal indexing is dependent on the data layout - contiguous data does not support it, but chunked data does. 

Therefore we could add an  `__orthogonal_indexing__` property that returns `True` or `False` depending on the data layout.

davidhassell:
## Description
Add `__orthogonal_indexing__` flag to the `Dataset` class.

Closes #171 

## Before you get started

- [x] ☝ Create an issue](https://github.com/NCAS-CMS/pyfive/issues) to discuss what you are going to do

## Checklist

- [x] All tests pass


bnlawrence (approved):
This is consistent with the discussion

valeriupredoi (approved):
LGTM with many thanks! @davidhassell please merge at will 🍺

PR #189, opened 2026-02-03
Replace python-lzf with python-neo-lzf

### LINKED ISSUE #144: python-lzf deprecation warning
With the new lzw support we have a dependency in testing (and users will have in production) on python-lzw. Currently when building this gives a build warning:
```
 DEPRECATION: Building 'python-lzf' using the legacy setup.py bdist_wheel mechanism, which will be removed in a future 
version. pip 25.3 will enforce this behaviour change. A possible replacement is to use the standardized build interface by 
setting the `--use-pep517` option, (possibly combined with `--no-build-isolation`), or adding a `pyproject.toml` file to the 
source tree of 'python-lzf'. Discussion can be found at https://github.com/pypa/pip/issues/6334
```
@valeriupredoi Is this something we need to do something about? Can we do something about it?

valeriupredoi:
## Description

Closes #144 

Thanks @bmaranville and @kmuehlbauer - I have moved from the old `python-lzf` to the newly built package (with constant wheels) `python-neo-lzf` from https://pypi.org/project/python-neo-lzf/ that also has a conda forge candidate.

## Before you get started

## Checklist

- [x] All tests pass

### DIFF
--- a/pyfive/btree.py
+++ b/pyfive/btree.py
@@ -1,6 +1,7 @@
 """HDF5 B-Trees and contents."""
 
 from collections import OrderedDict
+import lz4.frame as lzf
 import struct
 import zlib
 
@@ -186,12 +187,6 @@ class BTreeV1RawDataChunks(BTreeV1):
                 # strip off 4-byte checksum from end of buffer
                 chunk_buffer = chunk_buffer[:-4]
             elif filter_id == LZF_FILTER:
-                try:
-                    import lzf
-                except ImportError as e:
-                    raise ModuleNotFoundError(
-                        "LZF codec requires optional package 'python-lzf'."
-                    ) from e
                 uncompressed_len = struct.unpack(">H", chunk_buffer[:2])[0]
                 chunk_buffer = lzf.decompress(chunk_buffer, uncompressed_len)
             else:
--- a/pyfive/btree.py
+++ b/pyfive/btree.py
@@ -1,7 +1,6 @@
 """HDF5 B-Trees and contents."""
 
 from collections import OrderedDict
-import lz4.frame as lzf
 import struct
 import zlib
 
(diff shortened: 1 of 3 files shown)


pyfive/btree.py
```diff
@@ -1,6 +1,7 @@
 """HDF5 B-Trees and contents."""
 
 from collections import OrderedDict
+import lz4.frame as lzf
```
bmaranville: The LZF filter is registered with id 32000, and the LZ4 filter is registered with id 32004 - I don't think they're interchangeable.

valeriupredoi: you're absolutely right @bmaranville - big fail on my part for trusting ChatGPT saying that lz4 can decompress lzf :grin:

valeriupredoi: resolving this since lz4 is completely abandoned now

pyfive/btree.py
```diff
@@ -187,6 +186,12 @@ def _filter_chunk(cls, chunk_buffer, filter_mask, filter_pipeline, itemsize):
                 # strip off 4-byte checksum from end of buffer
                 chunk_buffer = chunk_buffer[:-4]
             elif filter_id == LZF_FILTER:
+                try:
+                    import lzf
+                except ImportError as e:
+                    raise ModuleNotFoundError(
+                        "LZF codec requires optional package 'python-lzf'."
```
bmaranville: Could this be `LZF codec requires optional conda package 'python-lzf' or pip package 'python-neo-lzf'`?

Just because people won't generally be able to install `python-lzf` with pip anymore.

valeriupredoi: conda package `python-neo-lzf` actually - a great catch, many thanks @bmaranville :beer:

valeriupredoi: done in 42dbf93bc58a8ca11e4c8788ebceb12d9fdbb6dd

kmuehlbauer: I think it is `python-lzf` for conda and `python-neo-lzf` for PyPI. We could ask the maintainers if they would engage with PyPI to release with the original name. Not sure if or how this might work.

valeriupredoi: no, the brand new and very timely and useful python-neo-lzf has sprung on conda-forge @kmuehlbauer https://anaconda.org/channels/conda-forge/packages/python-neo-lzf/overview

valeriupredoi: they're not able to release with the same name unless they remove the PyPI package first, which is something they can't do without the approval of the original developer

valeriupredoi:
@kmuehlbauer @bmaranville gents, I'd like to merge this to include it in a new tagged release, you folks concur? :beers:

PR #191, opened 2026-02-04
Timing + Optimisations + Minor fixes

### LINKED ISSUE #153: b-tree read performance still not good enough

### LINKED ISSUE #190: Need to be able to get some logging information about timing
Ideally we need to have some logging available about how long things take so we can optimise things in the future.
To that end, we need logging.info to time at least the b-tree reads, and obtaining a full set of attributes.

bnlawrence:
## Description

This was  primarily about delivering some lightweight logging of timings to aid object store optimisation.
Closes #190 

However, in doing that, it also dealt with one minor bug and some minor pylance complaints following the introduction of type hints. It also fixed a lot of performance issues that were shown up by the presence of logging.
In particular
 - there was a lot re-reading of attributes, some of which is now avoided, and some of which is now defended against via caching the attributes on first read.
 - we have introduced a new file buffer for use when the file is on S3, which allows us to cache file metadata on first read, regardless of the s3fs parameters. This vastly improves performance when opening and inspecting file contents, particularly if they have all their metadata properly at the front of the file.  

Closes #153 .

## Checklist

- [x] All tests pass


pyfive/dataobjects.py
```diff
@@ -7,6 +7,13 @@
 import warnings
 
 import numpy as np
+from time import time
+import logging
+import inspect
+try:
+    from importlib.metadata import version
```
valeriupredoi: can keep just this, Bryan, the module has been standard part of importlib since a long while now (incl Python 3.10 which we don't even support)

bnlawrence: You mean, just lose the try loop?

bnlawrence:
I think all of those failures are related to [this](https://github.com/NCAS-CMS/pyfive/issues/103#issuecomment-3859205718) ... I am not sure the current type checking is a good thing ...

pyfive/high_level.py
```diff
@@ -252,15 +272,34 @@ def __init__(self, filename: str | BinaryIO, mode: str = "r") -> None:
...
+            logger.info(
+                "[pyfive] Detected S3 file, enabling metadata buffering (%d MB)",
+                metadata_buffer_size,
+            )
+            self._fh = MetadataBufferingWrapper(fh, buffer_size=metadata_buffer_size)
+        else:
+            # Local file or other
+            # NOTE mypy detects incompatible types:
+            # str | BytesIO = MetadataBufferingWrapper
+            self._fh = fh  # type: ignore[assignment]
```
valeriupredoi: hey @bnlawrence all cleaned up now! Though please do have a look at this (and remove the NOTE if you OK with it) - I think mypy has a good point there, you're doing some interesting assignment gymnastics 😁

valeriupredoi (approved):
very cool indeed! Pls do have a look at my comment about fh assignment, and pls do resolve all the Copilot review comments, all are outdated 🍻

valeriupredoi:
just made it RfR for @davidhassell to have a looksee too :beer:

pyfive/utilities.py
```diff
@@ -39,3 +43,114 @@ def read(self, size=-1):
...
+    def __init__(self, fh, buffer_size: int = 1):
+        """
+        Initialize wrapper.
+
+        Parameters
+        ----------
+        fh : file-like
+            Original file handle (S3File or similar)
+        buffer_size : int
+            Size of metadata buffer to read upfront (MB, default=1MB)
```
davidhassell: :)
```suggestion
            Size of metadata buffer to read upfront (MiB, default=1MiB)
```

pyfive/high_level.py
```diff
@@ -230,6 +235,12 @@ class File(Group):
     filename : str or file-like
         Name of file (string or unicode) or file like object which has read
         and seek methods which behaved like a Python file object.
+    mode : str
+        File open mode (default: "r", read-only).
+    metadata_buffer_size : int
+        Size of metadata buffer for S3/remote files in MB (default: 1MB).
```
davidhassell: :)
```suggestion
        Size of metadata buffer for S3/remote files in MiB (default: 1MiB).
```

davidhassell (approved):
Looks very good! Learnt about `cast`. Couple of suggestions - ignore at will :)

PR #202, opened 2026-02-25
Add `typing_extensions` as core dependency and add install from conda GHA test

valeriupredoi:
## Description

W have to add `typing_extensions` as a core dependency since otherwise the package fails to import correctly, I have just rebuilt the conda package with that in mind at https://github.com/conda-forge/pyfive-feedstock/pull/25 and PyPI package will be updated from the next release onwards, after this gets merged. I also added a conda install GHA test.

## Before you get started

## Checklist

- [x] All tests pass

### DIFF
--- a/.github/workflows/install-from-conda.yml
+++ b/.github/workflows/install-from-conda.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - add_core_dep_and_conda_GHA
   schedule:
     - cron: "0 4 * * *"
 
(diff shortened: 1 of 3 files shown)


.github/workflows/install-from-conda.yml
```diff
@@ -0,0 +1,51 @@
+name: Install from conda-forge
+permissions: {}
+on:
+  push:
+    branches:
+      - main
+      - add_core_dep_and_conda_GHA
```
valeriupredoi:
```suggestion
```

PR #231, opened 2026-05-20
Issue 230 changes; addressing the possibility of corrupt data resetti…

### LINKED ISSUE #230: Handle invalid HDF5 files gracefully.
We have come across some data where some missing fields have been flushed to a file with invalid output. An example of this occurs in the NCAS O12 model output, where a p5dump of the file creates two versions of a dimension:

```
dimensions:
        nvertex = 4;
        y = 3606;
        x = 4322;
        time_counter = 1;
        axis_nbounds = 2;
        time_counter = 0;
        deptht = 75;
```

The second version arises from an empty field in the file, that ought not be empty:
```
DATASET "berg_latent_heat_flux" {
      DATATYPE  H5T_IEEE_F32LE
      DATASPACE  SIMPLE { ( 0, 3606, 4322 ) / ( H5S_UNLIMITED, 3606, 4322 ) }
      ATTRIBUTE "DIMENSION_LIST" {
         DATATYPE  H5T_VLEN { H5T_REFERENCE { H5T_STD_REF_OBJECT } }
         DATASPACE  SIMPLE { ( 3 ) / ( 3 ) }
      } ...
```
but in this case, the `time_counter` dimension is being set 0 here, even though it is 1 elsewhere.

Odd things happen, for example, h5py thinks there is no data in this variable, and netcdf4-python thinks there is.  (There is not).

AI seems to think this situation arises from some combination of
- The file write was interrupted (power loss, crash, killed process)
- The NetCDF library cached metadata that wasn't flushed to disk
- Buffer overflow or corruption during a large write operation

So, while this sort of file is invalid, most of the data in the file is fine, and pyfive should handle this gracefully.

bnlawrence:
This addresses, and closes #230, where data has been seen which is partially corrupt, possibly due to misconfigured libraries in upstream data generators. Since that's an issue which could recur, this pull request provides support within pyfive so that this situation can be identified, and the consequences mitigated.

## Checklist

- [ ] All tests pass

### DIFF
--- a/pyfive/high_level.py
+++ b/pyfive/high_level.py
@@ -6,6 +6,7 @@ from collections.abc import Callable
 from collections.abc import Mapping, Sequence
 from abc import ABC
 import os
+import warnings
 import posixpath
 import warnings
 import logging
@@ -601,8 +602,8 @@ class DimensionManager(Sequence):
         if "DIMENSION_LABELS" in dset.attrs:
             dim_labels = dset.attrs["DIMENSION_LABELS"]
         self._dims = [
-            DimensionProxy(dset.file, label, refs)
-            for label, refs in zip(dim_labels, dim_list)
+            DimensionProxy(dset, dset.file, label, refs, axis)
+            for axis, (label, refs) in enumerate(zip(dim_labels, dim_list))
         ]
 
     def __len__(self):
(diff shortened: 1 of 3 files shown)


valeriupredoi (approved):
this is very useful indeed @bnlawrence - could we have a couple real world integration tests with some really pathological files that are also not too big perhaps? 🍺

bnlawrence:
Short answer is I can't create data in this state, and the only example I have is 41 GB ...  all I could do was to mock the problem in testing, which is what i did.

valeriupredoi:
> Short answer is I can't create data in this state, and the only example I have is 41 GB ... all I could do was to mock the problem in testing, which is what i did.

ah completely understand the issue, Bryan - just for shits and giggles, would popping that 41G file in an S3 bucket reproduce the issue? Provided we don't load the data in memory

Issue #228, opened 2026-05-08
difference in check_dtype for vlen compared to h5py


kmuehlbauer:
`pyfive.check_dtype(vlen=var.dtype)` and `h5py.check_dtype(vlen=var.dtype)` return different. This fails in downstream `xarray` when using `engine="h5netcdf"` with `pyfive` backend.

In xarray check_dtype is used to check for vlen strings and decodes it from `object` to `U`. See below.

MCVE, pyfive/h5netcdf/xarray latest versions, h5py=13.5.1, hdf5=1.14.6

```python
import h5py
import pyfive
import xarray as xr
import os

input_string = ["foó", "bár", "baź"]
original = xr.Dataset({"x": input_string})

kwargs = dict(encoding={"x": {"dtype": str}})
fname = "test.nc"
original.to_netcdf(fname, engine="h5netcdf", **kwargs)

print("----- PYFIVE --------------------")
with pyfive.File("test.nc") as fh:
    var = fh["x"]
    print(pyfive.check_dtype(vlen=var.dtype))
    print(var.dtype.metadata)
    print(fh["x"][...])

print("\n----- H5PY --------------------")
with h5py.File("test.nc") as fh:
    var = fh["x"]
    print(h5py.check_dtype(vlen=var.dtype))
    print(var.dtype.metadata)
    print(fh["x"][...])


backend = "h5py"
os.environ["H5NETCDF_READ_BACKEND"] = backend
print(f"\n----- xarray - h5netcdf - {backend} --------------------")
with xr.open_dataset("test.nc", engine="h5netcdf") as ds:
    print(ds["x"])

backend = "pyfive"
os.environ["H5NETCDF_READ_BACKEND"] = backend
print(f"\n----- xarray - h5netcdf - {backend} --------------------")
with xr.open_dataset("test.nc", engine="h5netcdf") as ds:
    print(ds["x"])
```
```
----- PYFIVE --------------------
string_info(encoding='ascii', length=None)
{'vlen': <class 'str'>}
['foó' 'bár' 'baź']

----- H5PY --------------------
<class 'str'>
{'vlen': <class 'str'>}
[b'fo\xc3\xb3' b'b\xc3\xa1r' b'ba\xc5\xba']

----- xarray - h5netcdf - h5py --------------------
<xarray.DataArray 'x' (x: 3)> Size: 36B
array(['foó', 'bár', 'baź'], dtype='<U3')
Coordinates:
[...]

Issue #249, opened 2026-07-06
Chunk_buffer handling on timeouts


dwest77a:
Currently there's no handling for where a timeout occurs within aiohttp/fsspec, so the response object is interpreted as a bytes-like object on decompression. Ideally this can be caught further up the chain of events.

```
Traceback (most recent call last):
  File "/Users/daniel.westwood/Documents/CEDA_Documents/project_tracking_details/testing/file_access.py", line 87, in <module>
    raise exc
  File "/Users/daniel.westwood/Documents/CEDA_Documents/project_tracking_details/testing/file_access.py", line 84, in <module>
    s = future.result()
  File "/opt/homebrew/Cellar/python@3.13/3.13.9_1/Frameworks/Python.framework/Versions/3.13/lib/python3.13/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ~~~~~~~~~~~~~~~~~^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.9_1/Frameworks/Python.framework/Versions/3.13/lib/python3.13/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/opt/homebrew/Cellar/python@3.13/3.13.9_1/Frameworks/Python.framework/Versions/3.13/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/daniel.westwood/Documents/CEDA_Documents/project_tracking_details/testing/file_access.py", line 25, in _iterate_range
    return ds[i*multiplier:(i+1)*multiplier]  # np.min(ds[0:i])
           ~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/daniel.westwood/Documents/CEDA_Documents/project_tracking_details/testing/.venv/lib/python3.13/site-packages/pyfive/high_level.py", line 445, in __getitem__
    data = self.id.get_data(args, self.fillvalue)
  File "/Users/daniel.westwood/Documents/CEDA_Documents/project_tracking_details/testing/.venv/lib/python3.13/site-packages/pyfive/h5d.py", line 425, in get_data
    return self._get_selection_via_chunks(args)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
[...]

PR #258, opened 2026-08-27
Issue252 - dealing with user blocks in front of HDF5 superblocks

### LINKED ISSUE #252: Support for Matlab HDF5 files?
I am writing a QGIS plugin, and am trying to switch from using h5py to (vendored) pyfive since I have found that the biggest obstacle to user adoption is just getting h5py installed correctly.

One of the datasets that I need to support provides their data in .mat files; for earlier versions, I use `scipy.io` for them, but for later versions, that library recommends using a HDF5 reader.

The data can be downloaded from: 
https://data.cresis.ku.edu/data/rds/2022_Antarctica_BaslerMKB/CSARP_standard/20230106_01/

I tested with the [Data_20230106_01_014.mat](https://data.cresis.ku.edu/data/rds/2022_Antarctica_BaslerMKB/CSARP_standard/20230106_01/Data_20230106_01_014.mat) file

I then try to read it using:
```
import pyfive
filepath = "/path/to/datadir/Data_20230106_01_014.mat"
data = pyfive.File(filepath)
```
and I get the error:
```
  File "<input>", line 1, in <module>
  File "/Users/lindzey/Documents/QIceRadar/code/qiceradar_plugin/external/pyfive/pyfive/high_level.py", line 300, in __init__
    self._superblock = SuperBlock(self._fh, 0)
                       ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lindzey/Documents/QIceRadar/code/qiceradar_plugin/external/pyfive/pyfive/misc_low_level.py", line 40, in __init__
    raise NotImplementedError(
NotImplementedError: unsupported superblock version: 46
```

By contrast, this works:
```
import h5py
data = h5py.File(filepath)
```

The issue seems to be that Matlab-generated HDF5 files have a header before the HDF5 header, but pyfive only looks for the header at offset 0. 

I have a patch that is working for me, but I did not open a PR because it was generated by Claude and I think it's an ugly hack that's not how you'd actually want to solve this problem in your codebase, and I don't have the bandwidth to figure out the right/clean way and shepherd it through review.

bnlawrence:
## Description

#252 exposed a problem with our superblock handling. The HDF5 spec allows a user block in front of the HDF5 itself, and HDF5 implementations should look for the superblock at 0, and then at powers of 2 greater than 512 bytes.  This means that if we don't find a superblock at offset zero, we have to go looking for where it might be before rejecting the file as not being an HDF5 file.

Closes #252, #249 and #228.

## Checklist

- [ ] All tests pass

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
     rev: v6.0.0
     hooks:
       - id: check-added-large-files
-        exclude: ^tests/data/cmip_bad_eg\.nc$
+        exclude: ^tests/data/(cmip_bad_eg\.nc|Data_20230106_01_014\.mat)$
       - id: check-ast
       - id: check-case-conflict
       - id: check-merge-conflict
--- a/pyfive/dataobjects.py
+++ b/pyfive/dataobjects.py
@@ -56,7 +56,7 @@ class DataObjects(object):
     HDF5 DataObjects.
     """
 
-    def __init__(self, fh, offset, order="C"):
+    def __init__(self, fh, offset, order="C", decode_strings=False):
         """initalize."""
         # Log file handle info for diagnostics
         fh_id = id(fh)
@@ -96,6 +96,7 @@ class DataObjects(object):
         self._chunk_address = None
         self._cached_attributes = None  # Cache parsed attributes
         self.order = order
+        self.decode_strings = decode_strings
 
     @staticmethod
     def _parse_v1_objects(fh):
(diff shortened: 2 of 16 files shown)


davidhassell (approved):
Hi Bryan,

Nice job. All makes sense, and I can see that it optimised in the "not constantly adding zero when there's no user block" sense.

The test is simple but effective.

#### decode_vlen

This also looks OK, in terms of the tests look sensible and pass, and the core logic makes sense.

Should we also not the in the PR that this fixes #228?

PR #260, opened 2026-08-28
Fix reading global heap

davidhassell:
Closed in favour of #261
----

## Fix reading global heap

Closes issue 259 

No unit test has been provided, because I found it hard to recreate the conditions in a reliable way! However, the two use cases from the issue now work, and all existing tests pass, so I was going to leave it at that. Can try harder to make a units test, if you like ...

## Before you get started

## Checklist

- [x] All tests pass

### DIFF
--- a/pyfive/misc_low_level.py
+++ b/pyfive/misc_low_level.py
@@ -158,15 +158,22 @@ class GlobalHeap(object):
         if self._objects is None:
             self._objects = OrderedDict()
             offset = 0
-            while offset < len(self.heap_data):
+            # Note: Need to add 16 to the offset to ensure at least 16
+            #       bytes remain in the buffer to unpack a complete
+            #       GLOBAL_HEAP_OBJECT header, preventing struct.error
+            #       on trailing padding (< 16B). See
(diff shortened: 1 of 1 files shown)


pyfive/misc_low_level.py
```diff
@@ -158,15 +158,22 @@ def objects(self):
         if self._objects is None:
             self._objects = OrderedDict()
             offset = 0
-            while offset < len(self.heap_data):
+            # Note: Need to add 16 to the offset to ensure at least 16
+            #       bytes remain in the buffer to unpack a complete
+            #       GLOBAL_HEAP_OBJECT header, preventing struct.error
+            #       on trailing padding (< 16B). See
+            #       https://github.com/NCAS-CMS/pyfive/issues/259
+            while offset + 16 <= len(self.heap_data):
```
zequihg50: Where does the 16 come from? In the specification for a global heap object it says:

```
Object Size: This is the size of the object data stored for the object. The actual storage space allocated for the object data is rounded up to a multiple of eight.
```

Perhaps for this particular issue the object size was rounded to 16 but we might be missing the more general behavior. Probably this was the LLM guess based on the actual error.

PR #265, opened 2026-09-03
Issue264 - Fixing non-standard shuffle/filter order operations

### LINKED ISSUE #264: Cannot read SWIFT files due to filter pipeline
Thanks for releasing this, which looks of great interest for the pynbody project (https://github.com/pynbody/pynbody). It profiles well for using multiple threads to read concurrently (which libhdf5 prohibits).

However, at the moment, we can't adopt as it doesn't work for hdf files created with SWIFT (https://github.com/SWIFTSIM/SWIFT). This appears to be due to custom filter ordering in SWIFT. SWIFT writes fletcher32 -> shuffle -> deflate which pyfive appears to baulk at. 

Note that this may be related to the closed issue #140, but I don't understand the exact relationship.

Minimal example to reproduce:

```python
import os
import h5py
import numpy as np
import pyfive

PATH = "pyfive_filter_order.h5"
a = np.arange(512, dtype="<i8")

# build the pipeline in SWIFT's order via the low-level API
dcpl = h5py.h5p.create(h5py.h5p.DATASET_CREATE)
dcpl.set_chunk((512,))
dcpl.set_fletcher32()
dcpl.set_shuffle()
dcpl.set_deflate(4)

fid = h5py.h5f.create(PATH.encode(), h5py.h5f.ACC_TRUNC)
dsid = h5py.h5d.create(fid, b"d", h5py.h5t.py_create(a.dtype, logical=True),
                       h5py.h5s.create_simple((512,)), dcpl=dcpl)
dsid.write(h5py.h5s.ALL, h5py.h5s.ALL, a)
del dsid
fid.close()

with h5py.File(PATH, "r") as f:
    plist = f["d"].id.get_create_plist()
    print("filter order:", [plist.get_filter(i)[0] for i in range(plist.get_nfilters())],
          "(3=fletcher32, 2=shuffle, 1=deflate)")
    print("h5py  :", "OK" if np.array_equal(a, f["d"][...]) else "WRONG")

try:
    with pyfive.File(PATH) as f:
        print("pyfive:", "OK" if np.array_equal(a, f["d"][...]) else "WRONG")
except Exception as e:
    print(f"pyfive: {type(e).__name__}: {e}")

os.unlink(PATH)
```

bnlawrence:
## Description

This pull request handles non-standard shuffle/filter operations. While the original code did things in the right order, it turns out that different orders mean that the shuffle step is seeing a different length buffer. In this situation the chunk buffer reaching the shuffle-undo step isn't always an exact multiple of `itemsize`. The HDF5 C library handles this by shuffling only the largest itemsize-aligned prefix and passing any trailing remainder bytes through unchanged, which is now what we do.

Closes #264

## Checklist

- [ ] All tests pass

### DIFF
--- a/pyfive/btree.py
+++ b/pyfive/btree.py
@@ -283,22 +283,21 @@ class BTreeV1RawDataChunks(BTreeV1):
                 chunk_buffer = zlib.decompress(chunk_buffer)
             elif filter_id == SHUFFLE_FILTER:
                 buffer_size = len(chunk_buffer)
-                if buffer_size % itemsize != 0:
-                    if (buffer_size - 4) % itemsize == 0:
-                        chunk_buffer = chunk_buffer[:-4]
-                        buffer_size = len(chunk_buffer)
-                    else:
-                        raise ValueError(
-                            "shuffle filter length mismatch: "
-                            f"len={buffer_size}, itemsize={itemsize}"
-                        )
-                unshuffled_buffer = bytearray(buffer_size)
-                step = buffer_size // itemsize
+                # The HDF5 shuffle filter only operates on the largest prefix
+                # of the buffer that is an exact multiple of itemsize; any
+                # trailing remainder bytes (e.g. from a filter applied before
+                # shuffle in the pipeline, such as fletcher32) are left
+                # unshuffled and simply copied through unchanged.
+                remainder = buffer_size % itemsize
+                main_size = buffer_size - remainder
+                tail = chunk_buffer[main_size:]
+                unshuffled_buffer = bytearray(main_size)
+                step = main_size // itemsize
                 for j in range(itemsize):
                     start = j * step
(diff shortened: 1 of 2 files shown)


davidhassell (approved):
Looks good. I confirm that the new test fails without this PR, and passes with it.

The new line  in `btree.py`:
```python
                chunk_buffer = bytes(unshuffled_buffer) + tail
```
looked a bit odd (why convert the `bytarray` to `bytes`)? But it all make sense now that I've looked at it a bit. Whilst you could do something like `chunk_buffer = unshuffled_buffer; chunk_buffer.extend(bytarray(tail)`, that could have unforeseen problems downstream (things expecting bytes rather than bytearray), and any performance improvement is theoretical at best (given that the chunk_buffer will end up in a np.frombuffer call anyway at some future time).

PR #266, opened 2026-09-03
Make Global Heaps cache on file not on dataobjects

### LINKED ISSUE #262: Global heap collections are read many more times than necessary
While analyzing how many S3 GET requests pyfive makes when reading NASA GEDI HDF5 files, I noticed that a large share of the requests were for global heap collections (`GCOL`), and the same collection was being fetched many times over. One of the files I used has 217 distinct `GCOL` blocks but pyfive fetched them a total of 4,598 times - about 21 times each on average.

Global heap collections is where the values of variable-length attributes and datasets are stored. The important thing to note is that one GCOL's holdings are shared. The vlen values belonging to many different datasets are usually packed together into the same GCOL block. So when pyfive reads the attributes of one dataset, it loads a GCOL, and when it moves on to the next dataset, whose vlen attribute values happen to point into that same GCOL, it loads it again from the start.

I think the cause is simply the scope of the cache. pyfive keeps its global-heap cache in the `_global_heaps` dictionary that lives on each `DataObjects` / `DatasetID`. Because every dataset gets its own fresh cache, a GCOL that was already read for one dataset is never reused for the next one. When accessing many HDF5 datasets with vlen attributes, which is what my use cases were, repeated GCOL S3 requests grow into a large fraction of the total request count, and total runtime. 

My suggestion for a fix is to cache global heap collections once per file, keyed by the collection's address, and share that cache across every dataset and attribute read. Each GCOL would then be fetched from storage only once, no matter how many datasets reference it.

bnlawrence:
Optimise the extraction of global heaps to avoid repeating global heap loading for every variable.

## Description

Adds a test and makes the global heap cache resident on the file.

Closes #262 

## Checklist

- [ ] All tests pass

### DIFF
--- a/pyfive/dataobjects.py
+++ b/pyfive/dataobjects.py
@@ -57,7 +57,7 @@ class DataObjects(object):
     HDF5 DataObjects.
     """
 
-    def __init__(self, fh, offset, order="C", decode_strings=False):
+    def __init__(self, fh, offset, order="C", decode_strings=False, global_heaps=None):
         """initalize."""
         # Log file handle info for diagnostics
         actual_fh = unwrap_file_handle(fh)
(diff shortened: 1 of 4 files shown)


davidhassell (approved):
All good from my perspective (tests pass)

PR #267, opened 2026-09-03
Added more informative errors when trying to read netcdf3

### LINKED ISSUE #263: fail gracefully with netcdf3 files
The CMIP6 archive, and probably others, include a proportion of netcdf3 files, when folks are using pyfive to read this data, pyfive (correctly) fails, but it would be helpful if it failed with a "Can't read NetCDF3 files" rather than just "Not an HDF5 file" ... because folks do expect the CMIP6 data to be netCDF4/HDF5 ...

bnlawrence:
## Description

The title says it all 

Closes #263

## Checklist

- [ ] All tests pass

### DIFF
--- a/pyfive/misc_low_level.py
+++ b/pyfive/misc_low_level.py
@@ -43,8 +43,6 @@ class SuperBlock(object):
             )
 
         # verify contents
-        if contents["format_signature"] != FORMAT_SIGNATURE:
-            raise InvalidHDF5File("Incorrect file signature")
         if contents["offset_size"] != 8 or contents["length_size"] != 8:
             raise NotImplementedError("File uses none 64-bit addressing")
         self.version = contents["superblock_version"]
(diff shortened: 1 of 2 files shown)


pyfive/misc_low_level.py
```diff
@@ -43,8 +43,6 @@ def __init__(self, fh, offset):
             )
 
         # verify contents
-        if contents["format_signature"] != FORMAT_SIGNATURE:
-            raise InvalidHDF5File("Incorrect file signature")
```
kmuehlbauer, line 47: What happens if someone tries to open some other file (eg txt file)? Will there still be an informative error?

davidhassell (approved):
All good. Tested on the command line with various files (text, PP, netCDF3, HDF5)

pyfive/misc_low_level.py
bnlawrence: The superblock is only looked for after the high-level has confirmed it is an HDF5 file.

kmuehlbauer: Thanks, I got a bit confused here :grimacing:.

bnlawrence: (So, yes)

